PR #19, opened 2019-09-17
Conditional skips

philastrophist:
Summary of propsed changes:

Allow skipping of questions based on the state of the answers dictionary when asking from a `form`. The `default` of each question is used when it is skipped (`checkbox` uses checked options when `default` is not available. Tests included

        >>> f = form(q1=questionary.confirm("Hello?"),
                      q2=questionary.checkbox("World?",
                                         choices=[
                                             Choice("foo", checked=True),
                                             "bar"
                                         ],
                                         )).skip_if(q2=lambda x: not x['q1'])

        >>> result = f.ask()
        Hello? (Y/n )n
        >>> result['q1'] == False, and result['q2'] =  ("foo",)
        True

<s>However, this breaks python 3.5 since orderddicts are not used in kwargs. Either we fix that by doing some dependency tracking or we stop supporting 3.5.</s> (fixed)

### DIFF
--- a/questionary/form.py
+++ b/questionary/form.py
@@ -39,7 +39,7 @@ class Form:
         while not all(f.key in answers for f in self.form_fields):
             for field in self.form_fields:
                 if field.key in answers:
-                    def skip(): return True
+                    def skip(x): return True
                 else:
                     skip = self.skip_conditions.get(field.key, lambda x: False)
                 values = {f.question: answers[f.key] for f in self.form_fields
--- a/questionary/form.py
+++ b/questionary/form.py
@@ -15,6 +15,10 @@ def form(**kwargs: Question):
     return Form(*(FormField(k, q) for k, q in kwargs.items()))
 
 
+class DependencyError(Exception):
+    pass
+
+
 class Form:
     """Multi question prompts. Questions are asked one after another.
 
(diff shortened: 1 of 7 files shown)


philastrophist:
Update to propsed changes:

Can now do the same `skip_if` as before, except with more natural seeming comparisons. This should deal with all frequent use cases for skipping. If there is a situation which is not covered by the usual dunder methods, you can still use a function.
Tests included

        >>> q1=questionary.select("Hello?", [Choice('a', 1), 'b'])
        >>> q2=questionary.checkbox("World?",
                                         choices=[
                                             Choice("foo", checked=True),
                                             "bar"
                                         ],
                                         ))
        >>> f = form(q1=q1, q2=q2).skip_if(q2=q1>0)
        >>> result = f.ask()
        Hello? (Y/n )n
        >>> result['q1'] == 1, and result['q2'] =  ("foo",)
        True

tmbo:
Sorry, I wasn't able to review this yet - I was in the us for a conference. I like the idea and I'll take a look at the implementation over the weekend (same for your other PR)

philastrophist:
I've fixed the dependency problem with a `while` loop (no need to drop 3.5 support now). It works of course, but I'm not sure if its messy or not.

tmbo:
Thinking about this back and forth I don't think it is necessary to include this. I think it is easier and cleaner to just use a normal `if` in python and ask the questions one after another. I don't see a huge benefit in having this compared to the added complexity. 

Please let me know if you think there is a strong reason to go for this.

Issue #20, opened 2019-10-05
New question type: Autocomplete


joaquingx:
I would like to add a new question type, or maybe is an enhancement to `text` question type. Basically I want to add the option to autocomplete text, in my opinion it has a lot of advantages:
* For a list with more than 15+ items, `select` or another question type would be cumbersome.
* If the prompt is used continously for the same person, i think that this option would be his best option.
etc

I was learning about `prompt_toolkit` and seems quite possible(https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#autocompletion). What do you think @tmbo? If you think it is a good idea im ready to contribute :)

tmbo:
Yes I think that would be a great addition 👍 could be an extension to text but if the code is different I’d rather go for a new type

joaquingx:
Hi :smile: , this weekend i've been learning more about prompt_toolkit and questionary. In this research i found that questionary is not using the last version of  prompt_toolkit 2(2.0.10), and this might be an issue. Basically, in 2.0.7, autocompletion base class does not allow `formatted_text` in `display`. When I was exploring in prompt_toolkit, I found this example: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/examples/prompts/auto-completion/colored-completions-with-formatted-text.py that i wanted to use as an inspiration to implement this feature. So, is possible to upgrade prompt_toolkit version?

tmbo:
Sure, feel free to update :+1:

tmbo:
implemented and merged ❤️  - will be part of 1.4.0 release

PR #21, opened 2019-10-08
Add autocomplete question type. 

joaquingx:
This PR implements autocomplete feature:

```
def ask_pystyle(**kwargs):
    # create the question object
    question = questionary.autocomplete(
        "Choose ant specie",
        choices=[
            'Camponotus pennsylvanicus',
            'Linepithema humile',
            'Eciton burchellii',
            "Atta colombica",
            'Polyergus lucidus',
            'Polyergus rufescens',
        ],
        **kwargs)

    # prompt the user for an answer
    return question.ask()
```

Retaled to #20


joaquingx:
Hey @tmbo, can you give me a review please :smile:

joaquingx:
How its going @tmbo ?

yardensachs:
Should this be a feature in select instead of a different type?

tmbo:
yes that is actually a good idea. I guess it comes down to the use case: 

Is this meant to just suggest entries, but allow arbitrary user input ? -> if so should merge with text

OR

Is this more a better UI for a list with lots of items where you are supposed to select one (like a select, but with lots of items)? --> if so should be separate component

joaquingx:
@tmbo Actually, it allows arbitrary user input. Also, with a proper validation(https://github.com/tmbo/questionary/pull/21/files#diff-9839c6f9a625f87aef5dc23f53bb6e14R76) it will only allows input that are in `choices`. At first sight i think is possible to merge autocomplete with text but it will increase text complexity.

PR #30, opened 2019-12-17
MANIFEST.in: Update

jayvdb:
Include docs, examples and tests

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,6 @@
-include LICENSE README.md requirements.txt
+include LICENSE NOTICE README.md requirements.txt
+include pyproject.toml
+recursive-include docs *.gif
+recursive-include docs *.png
+recursive-include examples *.py
+recursive-include tests *.py
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
         "Programming Language :: Python :: 3.7",
         "Topic :: Software Development :: Libraries",
     ],
-    packages=find_packages(exclude=["tests", "examples"]),
+    packages=find_packages(exclude=["tests", "tests.*", "examples"]),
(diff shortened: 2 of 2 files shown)


tmbo:
Thanks a lot for your PR ⭐️ - Can you explain a bit why you added these additional files, why do we need to ship the docs / tests / examples as part of the package?

jayvdb:
It isnt necessary, but it helps distro packagers.  Especially the tests, as it allows the packagers to easily grab the tests for each version, and run them.

We can workaround this by fetching from GitHub, but that adds a step to the update process, as the release notice comes from PyPI, and then we need to fetch from GitHub hope the tag name convention doesnt change, and check that the contents of Github are a superset of the contents of PyPI - i.e. they have the same package code, otherwise distro users are getting a slightly different version than PyPI users.

The setup.py commit is needed irrespective, as it also fixes `pip install git+https://github.com/tmbo/questionary`

e.g. https://build.opensuse.org/package/view_file/home:jayvdb:py-new/python-questionary/python-questionary.spec?expand=1

`NOTICE` should be included as it is part of the license.

`pyproject.toml` helps distro maintainers check that they are bumping minimum versions of dependencies as needed.

`README.md` refers to images, so it is nice if they are included, but it is the least important.

The usual objection is that these docs, tests, etc increase the size of the sdist.  As you are already shipping wheels for this package, which most users will be fetching, the main use of the sdist will be people who have explicitly choosen to not get the wheel, and dont want 'code + license only' (i.e. the wheel) - they want the source.

tmbo:
That sounds very reasonable - I don't have any objections against including these files, just wanted to know the use case to keep this in mind in the future. 

Thanks a lot for this packaging improvement ⭐️

Issue #34, opened 2020-01-29
Improved documentation


xmatthias:
I think this is a great library - however documentation is somewhat lacking.

I've been able to build up a rather nice workflow now - but most of it i had to piece together from the sourcecode.

Mainly, it's about the configuration dictionary way:

```python
from questionary import Separator, prompt
questions = [
        {
            "type": "confirm",
            "name": "conditional_step",
            "message": "Would you like the next question?",
            "default": True,
        },
       {
            "type": "text",
            "name": "next_question",
            "message": "Name this library?",
            "when": lambda x: x['conditional_step'],
            "validate": lambda val: val == "questionary"
        },
       {
            "type": "select",
            "name": "second_question",
            "message": "Select item",
            "choices": [
                "item1",
                "item2",
                Separator(),
                "other",
            ],
        },
        {
            "type": "text",
            "name": "second_question",
            "message": "Insert free text",
            "when": lambda x: x["second_question"] == "other"
        },
]
prompt(questions)
```

For the above dictionary - the following points are missing / non-obvious in the docs
* `"validate"`: key name when used as dictionary
* `"validate": can be a function, does not have to be a ValidatorClass
* `"when"`: conditional - the readme.md only points this out as "skipif"
* `"second_question"`: Names can be reused in this way, to offer "choices" - or a "other" - free text form - which will end in the same dictionary key - "second_question" in the above example.

While the examples cover some of this - i think it would be better to have these very cool and important features documented properly.

tmbo:
@xmatthias that is great feedback, thanks a lot. Given that your experience is still fresh, do you mind proposing the changes to the Readme?

Maybe pointing towards the examples more could help (and adding more comments in there)

xmatthias:
I guess i'll close this as #38 has been merged - and it includes what i personally was missing (not sure if other things are missing too).

PR #38, opened 2020-01-30
Enhance documentation regarding advanced workflows

xmatthias:
As said in #34, the documentation lacks examples for advanced workflows, so this aims to provide some (for sure not complete ...).

I've also added a working example workflow - but only as dictionary configuration (`ask_dictstyle()`) for now.
Mainly because I'm not certain if this workflow style would also work for the pystyle calling possibilities without adding an if-statement and calling the questions one by one (as done in `examples/readme.py`).

The disadvantage this "one by one" calling (as done in `examples/readme.py` ) is that a user cannot easily interrupt the process with ctrl+c - but has to do that for every following question.

With the dict-style (and calling prompt only once), one interrupt will return the user immediately, no matter how many questions would still be left.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ Optional Keys:
 * `'choices'` - List of choices (applies when `'type': 'select'`) or function returning a list of choices.
 * `'when'` - Function checking if this question should be shown or skipped (same functionality than `.skip_if()`).
 * `'validate'` - Function or Validator Class performing validation (will be performed in real time as users type).
+* `filter` - Receive the user input and return the filtered value to be used inside the program. 
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -220,38 +220,38 @@ Questionary allows creating quite complex workflows when combining all of the ab
 ``` python
 from questionary import Separator, prompt
 questions = [
-        {
-            'type': 'confirm',
-            'name': 'conditional_step',
-            'message': 'Would you like the next question?',
-            'default': True,
-        },
-        {
-            'type': 'text',
-            'name': 'next_question',
-            'message': 'Name this library?',
-            # Validate if the first question was answered with yes or no
-            'when': lambda x: x['conditional_step'],
-            # Only accept questionary as answer
-            'validate': lambda val: val == 'questionary'
-        },
-        {
-            'type': 'select',
-            'name': 'second_question',
-            'message': 'Select item',
-            'choices': [
-                'item1',
-                'item2',
-                Separator(),
-                'other',
-            ],
-        },
-        {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -196,6 +196,89 @@ answers = prompt(questions)
...
+* `'name'` - The name of the question (will be used as key in the `answers` dictionary)
+* `'message'` - Message that will be shown to the user
+
+Optional Keys:
+
+* `'qmark'` - Question mark to use - defaults to `?`.
+* `'default'` - Preselected value.
+* `'choices'` - List of choices (applies when `'type': 'select'`).
+* `'when'` - Function checking if this question should be shown or skipped (same functionality than `.skip_if()`).
+* `'validate'` - Function or Validator Class performing validation (will be performed in real time as users type).
```
xmatthias: Please have a close look at this list - these are the keys i was able to figure out myself as I needed these for my usecase - but it may be possible that i missed one or more of these i'm not aware about.

tmbo:
```suggestion
* `'validate'` - Function or Validator Class performing validation (will be performed in real time as users type).
* `filter` - Receive the user input and return the filtered value to be used inside the program. 
```

README.md
```diff
@@ -196,6 +196,89 @@ answers = prompt(questions)
...
+
+</details>
+
+<details><summary>Advanced workflow examples</summary>
+Questionary allows creating quite complex workflows when combining all of the above concepts.
+
+``` python
+from questionary import Separator, prompt
+questions = [
+        {
```
tmbo: this should be indented by 4 spaces (currently 8)

README.md
```diff
@@ -196,6 +196,89 @@ answers = prompt(questions)
...
+            'when': lambda x: x['second_question'] == 'other'
+        },
+]
+prompt(questions)
+```
+
+The above workflow will show to the user as follows:
+1. Yes/No question `Would you like the next question?`.
+2. `Name this library?` - only shown when the first question is answered with yes
+3. Itemselection
```
tmbo:
```suggestion
3. A question to select an item from a list
```

tmbo:
Perfect, do you mind taking a look at the failing travis build - seems like there is a whitespace missing in the example you added

xmatthias:
I'm sorry...  of course i will.

I do look at CI results usually ... just give me some time ... i've fixed the other failure but didn't notice that the new commit also failed

I'll remember to keep this window open to check it fails (I get notifications for my repositories ...)

tmbo:
No worries 🙂 - looking all good, thanks a lot ⭐️

PR #48, opened 2020-06-26
Added validation for checkboxes

kiancross:
I've added validation to the checkbox prompt.

`prompt-toolkit` does not provide native validation for `FormattedTextControl` (the parent class of `InquirerControl`). Therefore while `text` prompts support validators in the form of a `Validator` subclass, this only accepts a `callable` returning a boolean and optional error message. This restriction means asynchronous validators supported by `prompt-toolkit` cannot be used for the checkbox.

If/when `prompt-toolkit` provides native validation for a `FormattedTextControl`, this validation logic for the checkbox can be replaced.

Even with its limitations, I think this covers the fast majority of use cases for checkbox validation - it should be useful to some people.

### DIFF
--- a/examples/checkbox_toppings.py
+++ b/examples/checkbox_toppings.py
@@ -4,7 +4,10 @@ import questionary
 if __name__ == "__main__":
     toppings = (
         questionary.checkbox(
-            "Select toppings", choices=["foo", "bar", "bazz"], style=custom_style_dope
+            "Select toppings",
+            choices=["foo", "bar", "bazz"],
+            validate=lambda a: (len(a) != 0, "You must select at least one topping"),
+            style=custom_style_dope,
         ).ask()
         or []
     )
--- a/questionary/prompts/checkbox.py
+++ b/questionary/prompts/checkbox.py
@@ -6,6 +6,7 @@ from prompt_toolkit.application import Application
 from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.styles import Style, merge_styles
+from prompt_toolkit.formatted_text import FormattedText
 
 from questionary.constants import DEFAULT_QUESTION_PREFIX, DEFAULT_STYLE
 from questionary.prompts import common
@@ -17,6 +18,7 @@ def checkbox(
     message: Text,
     choices: List[Union[Text, Choice, Dict[Text, Any]]],
     default: Optional[Text] = None,
+    validate: Any = lambda a: True,
     qmark: Text = DEFAULT_QUESTION_PREFIX,
     style: Optional[Style] = None,
     use_pointer: bool = True,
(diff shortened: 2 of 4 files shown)


nmstoker:
@tmbo what are your thoughts on this?

Seems to me that it would be a good feature to add validation on checkboxes, as without them it's feasible for users to accidentally "shoot past" a checkbox where _at least one_ selection is required if they've pressed return a load in succession.

In cases where the choice can be restricted to just one then clearly a select would handle this but where it's _one or more_ then the checkbox is needed and yet has the failing above.

examples/checkbox_toppings.py
```diff
@@ -4,7 +4,10 @@
 if __name__ == "__main__":
     toppings = (
         questionary.checkbox(
-            "Select toppings", choices=["foo", "bar", "bazz"], style=custom_style_dope
+            "Select toppings",
+            choices=["foo", "bar", "bazz"],
+            validate=lambda a: (len(a) != 0, "You must select at least one topping"),
```
tmbo, line 9: I think the API isn't ideal as there are three different types of arguments for the validate argument accross the different prompts:
- function returning one boolean
- function returning a tuple (predicated + error message)
- prompt toolkit validator 

Since it would be a bit more effort to unify this properly I am good with this for the moment.

tmbo (dismissed):
Looks good, can you please add this parameter to the documentation of the checkbox component in the README? Once that is done we are ready to go :rocket:

examples/checkbox_toppings.py
kiancross: @tmbo I'm happy to sort this before I merge. Do you think implementing it using the `Validator` from `prompt_toolkit` would be better?

tmbo: I don't think we can use the `Validator` from `prompt_toolkit`: its `validate` function takes a `document` of type `prompt_toolkit.document.Document` as its argument, which we don't have in this case. This seems we'd need to roll our own class (and in terms of backwards compatibility, support the prompt toolkit validator as well).

kiancross: @tmbo A document instance can be created as so, `prompt_toolkit.document.Document("Some text")`, which might be sufficient?

This PR would actually probably have been better submitted to https://github.com/prompt-toolkit/python-prompt-toolkit to fix https://github.com/prompt-toolkit/python-prompt-toolkit/issues/430, which we could then have used here. What would you prefer?

tmbo: I mean if upstream prompt toolkit is open to improve this on their end this would be even better 🙂 otherwise I think trying the `Document` creation might actually work, I didn't have time to take a closer look if there is something else happing when creating a `Document`, but if there are no side effects that prevent other functionality from working I think we are good

kiancross: @tmbo I will implement the `Document` method for now. If `prompt_toolkit` accept the change, then as we are using the `Document` object, the API for questionary will probably not change, so it would simply be a case of changing how it is implemented.

kiancross: @tmbo After looking at this more closely, I can't use a `Document` because it takes a string, and we need to pass an array to the validator.

If you are happy with it in its current form then I will merge - we can look at tidying up the interface later.

kiancross:
@tmbo Just checking if you are happy for this to be merged in in its current state? I did not end up making any changes to the API, as `Document` only takes a string, but we need to pass an array.

examples/checkbox_toppings.py
tmbo: yes fine with me :+1:

tmbo (approved):
awesome, thanks a lot for implementing this, makes the interface more complete :rocket:

PR #55, opened 2020-09-08
Fixed #45 (colour of answer for some prompts)

### LINKED ISSUE #45: Color of answer for text prompt doesn't change
I've just switched over from PyInquirer to questionary and it was a seamless transition.  Kudos to you for creating questionary.  One thing I noticed under PyInquirer and a text prompt is that when I get the answer, the answer is high lighted in a different color.  Under questionary, that color doesn't change.  Am I missing something?  I would think that the answer should be able to be styled as the other prompts but I can't get it working no matter what I do.

Thanks.

kiancross:
Fixed #45.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ jobs:
   include:
   - stage: codestyle
     python: 3.7
-    install: pip install black
+    install: pip install -r requirements_dev.txt
     script: black --check .
     after_success: skip
   - stage: deploy
--- a/questionary/prompts/autocomplete.py
+++ b/questionary/prompts/autocomplete.py
@@ -178,7 +178,7 @@ def autocomplete(
 
     p = PromptSession(
         get_prompt_tokens,
-        lexer=SimpleLexer('class:answer'),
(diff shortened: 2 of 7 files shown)


kiancross:
Had to add some extra commits to fix `black` code style check (they've updated the tool, so previously passing code was failing.

@tmbo Could you review?

tmbo:
> Also no idea why Travis CI - Pull Request Queued — Build Created is stuck on queued?
seems like travis missed to report the result of the build back to github...the build passed ✅

tmbo (approved):
looks great 🚀

tmbo:
thanks a lot for updating black as well :) can you please update the black version in `requirements-dev.txt` to make sure it stays consistent?

kiancross:
> thanks a lot for updating black as well :) can you please update the black version in `requirements-dev.txt` to make sure it stays consistent?

It looks the version was already set, but the travis file was installing `black` directly and not using the `dev_requirements.txt` file. Fixed now.

tmbo (approved):
awesome stuff 🚀

Issue #43, opened 2020-03-05
Programmatically define pointed-at option for select


beeb:
Let me preface this by saying that I love questionary and I'm thankful to you guys for making it :)

I would like to suggest adding a feature to the `select` prompt. Currently, I think (by looking at the code) that there is no way to change which item in the select is pointed at (has the pointer symbol) when it gets initialized. It's always the first selectable choice that gets pointed at.

I would like to be able to start the `select` in a state where the pointer is on another choice (for instance the first one that has a value matching the `default` argument of the `select` call).
Another option would be to have an additional argument for `Choice` objects where we can say we want it to be pointed at initially.

Passing the `default` argument highlights the Choice that corresponds, but the pointer is still in the first position.

Thanks!

tmbo:
@beeb I think that is a great idea - selecting the default on startup should be a reasonably easy thing to do :+1:

PR #56, opened 2020-09-09
Add `initial_choice` for select and checkbox prompts

kiancross:
Implements #43.

### DIFF
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -152,12 +152,11 @@ class InquirerControl(FormattedTextControl):
         self.use_pointer = use_pointer
         self.default = default
 
-        if default is not None:
-            if default not in choices:
-                raise ValueError(
-                    f"Invalid `default` value passed. The value (`{default}`) does not exist in "
-                    f"the set of choices. Please make sure the default value is one of the available choices."
-                )
+        if default is not None and default not in choices:
+            raise ValueError(
+                f"Invalid `default` value passed. The value (`{default}`) does not exist in "
+                f"the set of choices. Please make sure the default value is one of the available choices."
+            )
 
         if initial_choice is not None:
             if initial_choice in choices:
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -158,17 +158,15 @@ class InquirerControl(FormattedTextControl):
                 f"the set of choices. Please make sure the default value is one of the available choices."
             )
 
-        if initial_choice is not None:
-            if initial_choice in choices:
-                self.pointed_at = choices.index(initial_choice)
-            else:
-                raise ValueError(
-                    f"Invalid `initial_choice` value passed. The value (`{initial_choice}`) does not exist in "
-                    f"the set of choices. Please make sure the initial value is one of the available choices."
-                )
-
-        else:
+        if initial_choice is None:
             self.pointed_at = None
+        elif initial_choice in choices:
+            self.pointed_at = choices.index(initial_choice)
+        else:
+            raise ValueError(
+                f"Invalid `initial_choice` value passed. The value (`{initial_choice}`) does not exist in "
+                f"the set of choices. Please make sure the initial value is one of the available choices."
+            )
 
         self.is_answered = False
         self.choices = []
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -157,7 +157,8 @@ class InquirerControl(FormattedTextControl):
                 self.pointed_at = choices.index(initial_choice)
             else:
                 raise ValueError(
-                    "Invalid 'initial_choice' value: must exist in 'choices'."
+                    f"Invalid `initial_choice` value for passed, the value (`{initial_choice}`) does not exist in "
+                    f"the set of choices. Please make sure the initial value is one of the available choices."
                 )
 
         else:
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -143,6 +143,7 @@ class InquirerControl(FormattedTextControl):
         use_indicator: bool = True,
         use_shortcuts: bool = False,
         use_pointer: bool = True,
+        initial_choice: Optional[Union[Text, Choice, Dict[Text, Any]]] = None,
         **kwargs
     ):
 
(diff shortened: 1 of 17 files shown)


dmarkey:
Anything stopping this going in? I need this 👍

questionary/prompts/common.py
```diff
@@ -151,7 +152,17 @@ def __init__(
         self.use_pointer = use_pointer
         self.default = default
 
-        self.pointed_at = None
+        if initial_choice is not None:
+            if initial_choice in choices:
+                self.pointed_at = choices.index(initial_choice)
+            else:
+                raise ValueError(
+                    "Invalid 'initial_choice' value: must exist in 'choices'."
```
tmbo, line 160:
```suggestion
                    f"Invalid `initial_choice` value for passed, the value (`{initial_choice}`) does not exist in "
                    f"the set of choices. Please make sure the initial value is one of the available choices."
```

questionary/prompts/common.py
```diff
@@ -151,7 +152,17 @@ def __init__(
         self.use_pointer = use_pointer
         self.default = default
 
-        self.pointed_at = None
+        if initial_choice is not None:
+            if initial_choice in choices:
+                self.pointed_at = choices.index(initial_choice)
```
tmbo, line 157: do we need to check if this choice can actually be selected (as in: making sure it is not disabled)?

kiancross: @tmbo This is done here:

https://github.com/tmbo/questionary/pull/56/files#diff-0eccf09aa36dc0dd59d841a342a6b76bR175-R179

tmbo: :+1:

questionary/prompts/select.py
```diff
@@ -23,6 +23,7 @@ def select(
     use_indicator: bool = False,
     use_pointer: bool = True,
     instruction: Text = None,
+    initial_choice: Optional[Union[Text, Choice, Dict[Text, Any]]] = None,
```
Yajo, line 26: IMHO this should be `default`, just like in any other question.

kiancross: Doing that would mean `default` behaves differently on the `select` type and `checkbox` type. As I said in #57 (where discussion is already ongoing), I think it is better we leave `default` alone on the `checkbox` type (seeing as it isn't broken there), and remove it completely from `select` (seeing as it doesn't do anything).

We can then add this new parameter to both `select` and `checkbox`, which indicates the initial pointed to item.

Yajo: > Doing that would mean `default` behaves differently on the `select` type and `checkbox` type

Yes, that'd be the expected behavior. You also don't expect it to work the same way in a text input, etc.

Honestly I see no practical use in adding `initial_choice` in checkboxes, the same as I don't see a reason to add it here: just making `default` work as expected should be enough. :roll_eyes:

kiancross: > Yes, that'd be the expected behavior. You also don't expect it to work the same way in a text input, etc.

No, however `select` and `checkbox` are similar and it would be confusing to have parameters with the same name, that do different things.

> Honestly I see no practical use in adding `initial_choice` in checkboxes, the same as I don't see a reason to add it here

Many of the use cases for being able to point to an arbitrary initial value for the `select` type are also applicable to the `checkbox` type. I see no reason to restrict functionality here.

tmbo: I don't think there is a use case where you'd want to point to any other but the default item in a select, is there? If not, I'd be for fixing the behaviour of default and avoid adding another attribute

kiancross: @tmbo How do you think it should behave with `checkbox` if we do this? Remove the ability to select the position?

Yajo: > The question is whether this enhancement should be added to `checkbox` too

I think the answer is no.

The fact you hit <kbd>Enter</kbd> in a checkbox just sends the checkbox answer, but it's going to end up with the same result, no matter where the cursor is located.

So the initial position in a checkbox is mostly irrelevant, and the initial position in a select must be exactly the same as the default value, always without exception.

So I agree with @tmbo. I think that all we need to do is to fix the behavior of `default` in selects.

Yajo (approved):

Yajo:
I'm needing this fix so badly... all questions have been solved, so any chance to get this merged soon? :pray:

kiancross:
@Yajo Just waiting for @tmbo to review - I can merge once he's happy with the changes.

questionary/prompts/common.py
```diff
@@ -151,7 +152,25 @@ def __init__(
         self.use_pointer = use_pointer
         self.default = default
 
-        self.pointed_at = None
+        if default is not None:
+            if default not in choices:
+                raise ValueError(
+                    f"Invalid `default` value passed. The value (`{default}`) does not exist in "
+                    f"the set of choices. Please make sure the default value is one of the available choices."
+                )
```
tmbo, line 160:
```suggestion
        if default is not None and default not in choices:
            raise ValueError(
                f"Invalid `default` value passed. The value (`{default}`) does not exist in "
                f"the set of choices. Please make sure the default value is one of the available choices."
            )
```

questionary/prompts/common.py
```diff
@@ -151,7 +152,25 @@ def __init__(
...
+            if initial_choice in choices:
+                self.pointed_at = choices.index(initial_choice)
+            else:
+                raise ValueError(
+                    f"Invalid `initial_choice` value passed. The value (`{initial_choice}`) does not exist in "
+                    f"the set of choices. Please make sure the initial value is one of the available choices."
+                )
+
+        else:
+            self.pointed_at = None
```
tmbo, line 172:
```suggestion
        if initial_choice is None:
            self.pointed_at = None
        elif initial_choice in choices:
            self.pointed_at = choices.index(initial_choice)
        else:
            raise ValueError(
                f"Invalid `initial_choice` value passed. The value (`{initial_choice}`) does not exist in "
                f"the set of choices. Please make sure the initial value is one of the available choices."
            )
```

tmbo (approved):
looks great, added some optional code suggestions - but otherwise good to merge :star:

PR #58, opened 2020-09-24
Add py.typed marker file

layday:
Questionary has type annotations but in the
absence of `py.typed` the annotations might be
ignored or the types incorrectly inferred
by type checkers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
         "Topic :: Software Development :: Libraries",
     ],
     packages=find_packages(exclude=["tests", "tests.*", "examples"]),
+    package_data={"": ["py.typed"]},
     version=__version__,
     install_requires=install_requires,
     tests_require=tests_requires,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -37,6 +37,7 @@
         "Topic :: Software Development :: Libraries",
     ],
     packages=find_packages(exclude=["tests", "tests.*", "examples"]),
+    package_data={"": ["py.typed"]},
```
tmbo, line 40:
```suggestion
    package_data={"questionary": ["py.typed"]},
```
I think this should mention the package, shouldn't it?

tmbo (commented):
looks good :+1: thanks a lot for adding this

setup.py
layday: `""` means from any package in `packages`.  I omitted the name of the package because questionary is detected automatically by `find_packages` but I can change it to`"questionary"` if you prefer.

tmbo: ah great to know, I wasn't aware of that - let's make it explicit though 🙂

kiancross (approved):

layday:
Thanks for the quick review 🚀

PR #59, opened 2020-09-25
Document multiline input for devs and users

### LINKED ISSUE #54: Multiline input
I'd need an option to have multiline input.

Something similar to https://python-prompt-toolkit.readthedocs.io/en/stable/pages/asking_for_input.html#multiline-input

Is that possible or planned?

Yajo:
For devs, this feature, that was already working, is now documented.

For users, a default instruction text appears when using multiline mode.

Fix #54.

### DIFF
--- a/questionary/prompts/text.py
+++ b/questionary/prompts/text.py
@@ -7,7 +7,11 @@ from prompt_toolkit.shortcuts.prompt import PromptSession
 from prompt_toolkit.styles import Style, merge_styles
 from prompt_toolkit.lexers import SimpleLexer
 
-from questionary.constants import DEFAULT_QUESTION_PREFIX, DEFAULT_STYLE
+from questionary.constants import (
+    DEFAULT_QUESTION_PREFIX,
+    DEFAULT_STYLE,
+    INSTRUCTION_MULTILINE,
+)
 from questionary.prompts.common import build_validator
 from questionary.question import Question
 
@@ -18,6 +22,8 @@ def text(
     validate: Any = None,
     qmark: Text = DEFAULT_QUESTION_PREFIX,
     style: Optional[Style] = None,
+    multiline: bool = False,
+    instruction: Optional[Text] = None,
     **kwargs: Any
 ) -> Question:
     """Prompt the user to enter a free text message.
@@ -44,6 +50,11 @@ def text(
         style: A custom color and style for the question parts. You can
                configure colors as well as font types for different elements.
 
+        multiline: If `True`, multiline input will be enabled.
+
+        instruction: Write instructions for the user if needed. If `None`
+                     and `multiline=True`, some instructions will appear.
+
     Returns:
         Question: Question instance, ready to be prompted (using `.ask()`).
     """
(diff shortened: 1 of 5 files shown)


kiancross:
Thanks for the PR. There's a few tests failing and also you need to run `black .` over the code, to ensure the formatting is correct.

kiancross:
Thanks for fixing code style. Final thing is a couple of tests to make sure the following lines are covered:
https://coveralls.io/builds/33727110/source?filename=questionary%2Fprompts%2Ftext.py#L63

questionary/prompts/text.py
```diff
@@ -52,14 +59,21 @@ def text(
 
     validator = build_validator(validate)
 
+    if instruction is None and multiline:
+        instruction = "(Finish with 'Alt+Enter' or 'Esc, Enter')"
```
Yajo, line 63: How am I supposed to assert that a tty displays this message properly?

kiancross: Good question - I should have looked more closely at the lines being highlighted! Don't think a test is required in this case then.

tmbo, line 63: Can you please move this as a constant to `questionary.constants`?

questionary/prompts/text.py
```diff
@@ -18,6 +18,8 @@ def text(
     validate: Any = None,
     qmark: Text = DEFAULT_QUESTION_PREFIX,
     style: Optional[Style] = None,
+    multiline: bool = False,
```
tmbo, line 21: should this rather be its own type instead of a flag? e.g. `textarea` 

@kiancross what do you think?

tests/test_question.py
```diff
@@ -62,3 +62,21 @@ def test_async_ask_question():
         assert response == "World"
     finally:
         inp.close()
+
+
+def test_multiline_text():
```
tmbo, line 67: can we also add a test to check the other key combination to submit the input, `Alt+Enter`?

questionary/prompts/text.py
kiancross: @tmbo I think as types are defined by functions, and there is significant overlap between these two types, using a flag is better.

In an ideal world, I think the types would be implemented as classes, then someone like this could be done with minimal code repetition using inheritance.

kiancross: @tmbo Can you think of any other ways a multi-line input may diverge from a text input? If not, I think using the same type is ok.

tests/test_question.py
Yajo: [It seems it's the same](https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/key_bindings.html#binding-alt-something-option-something-or-meta-something).

Yajo:
I had to fix one problem on your tests: you were using `\n` as the Enter key, while it is `\r` instead. Also another problem in my test. It should go :heavy_check_mark: now.

questionary/prompts/text.py
Yajo: It uses a few more attributes as you can read in upstream docs, but personally I think it is ok as an attribute.

kiancross: Will leave this unresolved for @tmbo to comment.

Yajo:
So, could this be merged?

kiancross:
@Yajo Waiting to see what @tmbo thinks about leaving this as the same type vs creating a new one.

questionary/prompts/text.py
tmbo: sounds good to me, let's keep it an attribute then

kiancross (approved):

Yajo:
According to the new rules for Hacktoberfest 2020:

> Pull requests labelled as hacktoberfest-accepted will count toward to Hacktoberest from any public repository on GitHub

So may I ask you please to label this PR as hacktoberfest-accepted? :pray: :pleading_face:

kiancross:
>So may I ask you please to label this PR as hacktoberfest-accepted?

@tmbo Can you check this?

tmbo:
already added it to the repository :+1:

PR #60, opened 2020-09-27
Accept style tuples in `title` argument annotation of `Choice`

layday:
Passing a style tuple would produce a type error because `title` was
annotated to only take strings.

### DIFF
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -22,8 +22,8 @@ from questionary.constants import (
 # which does not exist in v2 of prompt_toolkit
 FormattedText = Union[
     Text,
-    List[Tuple[str, str]],
-    List[Tuple[str, str, Callable[[Any], None]]],
+    List[Tuple[Text, Text]],
+    List[Tuple[Text, Text, Callable[[Any], None]]],
     None,
 ]
 
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -10,7 +10,7 @@ from prompt_toolkit.layout import (
     Window,
 )
 from prompt_toolkit.validation import Validator, ValidationError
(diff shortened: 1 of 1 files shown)


questionary/prompts/common.py
```diff
@@ -10,21 +10,30 @@
...
 )
 
+# This is a cut-down version of `prompt_toolkit.formatted_text.AnyFormattedText`
+# which does not exist in v2 of prompt_toolkit
+FormattedText = Union[
+    Text,
+    List[Tuple[Text, Text]],
+    List[Tuple[Text, Text, Callable[[Any], None]]],
+    None,
+]
```
kiancross, line 28: We are using prompt_toolkit v3, so I think you can use `prompt_toolkit.formatted_text.AnyFormattedText`?

layday:
`Text` is redundant if the library does not support Python 2 but I changed the `str`s to `Text` for consistency.

questionary/prompts/common.py
layday: Questionary supports both versions from what I can tell.

kiancross: Ah yes you're right, my bad. Will let Tom review too, but looks good - thanks for the PR.

kiancross:
@tmbo Can you review?

layday:
Might also make sense to add type checking to CI while we're here, if you'd like.  I don't know if you use a type checker for development?

kiancross:
@layday Yes I think that would be good. Perhaps in a separate pull request as I think because type checking is not part of CI at the moment, bugs may have crept in. What do you think?

layday:
Opened #61.

tmbo (approved):
looks great :+1:

PR #63, opened 2020-10-04
Added mypy type checking

### LINKED ISSUE #61: Adopting a type checker
I wonder if there would be any interest in adopting a type checker.  Questionary appears to be extensively annotated so this shouldn't prove too onerous.   The four established type checkers are mypy, Pyright, pytype and pyre - would you have a preference for one of those?

tmbo:
Added mypy to validate type in CI. missing stubs for promp_toolkit will be ignored.

- fixes #61

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,11 @@ jobs:
     install: pip install -r requirements_dev.txt
     script: black --check --diff .
     after_success: skip
+  - stage: codestyle
+    python: 3.7
+    install: pip install -r requirements_dev.txt
+    script: mypy questionary
+    after_success: skip
   - stage: deploy
     python: 3.6
     install: skip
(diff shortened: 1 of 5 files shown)


kiancross (approved):

kiancross:
Looks fine to me

.travis.yml
```diff
@@ -27,6 +27,11 @@ jobs:
     install: pip install -r requirements_dev.txt
     script: black --check --diff .
     after_success: skip
+  - stage: codestyle
```
kiancross, line 30:
```suggestion
  - stage: mypy
```
Is it better to use a different stage name, or is it ok to have duplicates?

kiancross, line 30:
```suggestion
  - stage: mypy
```
Is it better to use a different stage name, or is it ok to have duplicates?

tmbo: using the same name will make sure travis groups them 🙂

PR #67, opened 2020-10-04
New build system

tmbo:
- switch to poetry build system (replacing pip & requirements files)
- switch to github actions for ci builds (replacing travis)
- make sure versions in different locations are the same (git tag, version.py, pyproject.toml)
- drop py 3.5 for black


kiancross:
@tmbo Not sure why, but the following had been running for over an hour:
https://github.com/tmbo/questionary/pull/67/checks?check_run_id=1206038543
https://github.com/tmbo/questionary/pull/67/checks?check_run_id=1206038561
https://github.com/tmbo/questionary/pull/67/checks?check_run_id=1206038573

It looks like Windows using `prompt_toolkit` v2 didn't work? It was stuck on the testing code stage. I cancelled them so it didn't use up too many of the free action minutes. Perhaps we should add a 10 minute timeout, in case of bugs such as this, as we know testing shouldn't take that long?

tmbo:
good point :+1: I've added a timeout and also excluded the tests that fail to complete

kiancross:
@tmbo Any idea why it was failing?

tmbo:
not really - not sure if it makes sense to investigate given that it is for prompt toolkit 2. but in any case should happen independently (as windows wasn't tested at all before switching to gh actions)

.github/workflows/continous-integration.yml
```diff
@@ -0,0 +1,123 @@
...
+    - name: Set up Python 3.7 🐍
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Install poetry 🦄
+      uses: Gr1N/setup-poetry@v3
+
+    - name: Build ⚒️ Distributions
+      run: poetry build -u __token__ -p ${{ secrets.PYPI_TOKEN }}
```
kiancross, line 123: Do you think it might be nicer to have a new GitHub action that is run every time a new release is created (and not on each pull request): https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release

This action would deploy the new release to PyPi

tmbo: well, we still want to make sure that tests pass before releasing. it wasn't possible to do that when github actions got released, not sure if they added something more recently to allow depending on other workflows

kiancross: I think if code is only ever added to `master` via a pull request, then it is guaranteed to have passed the tests. The flow would be something like this:

1. PR comes in.
2. Tests run.
3. If tests have passed then merge into master.
4. (No workflows are run now).
5. Create a release.
6. Run the deploy workflow to push to PyPi.

kiancross: That's a good point!

kiancross (approved):

PR #68, opened 2020-10-04
Increase type coverage

layday:
This PR adds annotations to attributes and some parameters which lacked them.  The types of attributes of `py.typed` packages cannot be inferred - public attributes have to be annotated.  (Arguably, some of these attributes should be private but I did not want to introduce any breaking changes.)  I have also replaced `Text` - a type which was introduced for Python 2 interop - with `str`, removed the Python 2-specific encoding headers and simplified `super` calls.


layday:
I suppose I haven't really explained why I'm concerned with annotating some of these not very forward-facing types.  In my own code, I've reimplemented `checkbox` and `select` to add a few [extra key bindings](https://github.com/layday/instawow/blob/main/instawow/prompts.py) and my type checker is not very pleased with Questionary's internals.  More generally, I think that composing new prompts with `Choice`, `InquirerControl` and `Question` is a valid use case - even when it isn't intended for working around limitations with existing controls - though it does imply having to be more cautious with making changes to these classes.

kiancross:
Thanks for the PR. Looks good. Will let @tmbo review too.

tmbo:
sounds like a great idea, lets go for it :rocket:

kiancross (approved):

PR #69, opened 2020-10-05
Install Python version from matrix

layday:
The tests would only run on 3.7.


kiancross:
Good spot!

tmbo (approved):
lol...thanks so much for catching this 🙏 , quite a copy and paste error from my end 😅

PR #72, opened 2020-10-05
Indicate the multiline input with \n>

Yajo:
This feels more natural because multiline prompts usually start at the beginning of the terminal somehow. In a long question, the answer could start too far in the right side, and next lines too.

Of course, it can be overriden with the `instruction` parameter of the `text` prompt, so I'm just supplying a saner default.


tmbo (approved):
simple but great improvement to the UX, love it 🚀

PR #79, opened 2020-10-16
Use more agnostic interface types

Yajo:
This allows to use a generator for `questions` because:

```python
>>> from typing import *
>>> issubclass(List, Iterable)
True
>>> issubclass(Generator, Iterable)
True
>>> issubclass(Generator, List)
False
```

Also allows to use other mappings, such as `collections.ChainMap`, for `answers`, because:

```python
>>> issubclass(ChainMap, Mapping)
True
>>> issubclass(Dict, Mapping)
True
>>> issubclass(ChainMap, Dict)
False
```

### DIFF
--- a/questionary/prompt.py
+++ b/questionary/prompt.py
@@ -25,9 +25,10 @@ def prompt(
     if isinstance(questions, dict):
         questions = [questions]
 
-    answers = answers or {}
+    answers = dict(answers or {})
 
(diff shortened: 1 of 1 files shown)


kiancross:
Looks ok to me. @tmbo can you review.

tmbo (approved):
thanks a lot, really cool :rocket

PR #86, opened 2020-10-31
Api changes

### LINKED ISSUE #44: Feature request: file selection dialog
It would be great to have a file selection prompt for questionary.

Here's something I put together in a hurry that uses the built-in `select` prompt, but it's nowhere near perfect. For instance, it's recursive which I would not do if implementing seriously.

```python
from typing import Optional, List, Union
from pathlib import Path

from questionary import Choice, Separator, select


def questionary_file_select(
    message: str = 'Select a file', starting_dir: Path = Path(), glob: str = '*.*'
) -> Optional[Path]:
    """Prompt the user to select a file from the filesystem using CLI.

    The function recursively asks for a file until one is selected, so it's possible to traverse the directory structure
    by selecting a folder instead (or the parent folder). A custom glob pattern for the files list can be set to filter
    by file extension for instance.

    Args:
        message (str, optional): Instruction message for the user. Defaults to 'Select a file'.
        starting_dir (Path, optional): Starting location for the file browsing. Defaults to
            ``Path()``.
        glob (str, optional): Glob pattern for files listing, e.g. to filter by extension. Defaults to ``'*.*'``.

    Returns:
        Optional[Path]: the user-select file as a Path object or None if the user cancels
    """
    choices: List = [Choice('Cancel'), Choice('..', value=starting_dir.parent)]

    dirs = filter(lambda x: x.is_dir(), sorted(starting_dir.glob('*')))
    for directory in dirs:
        choices.append(Choice(directory.stem, value=directory))

    choices.append(Separator())

    files = filter(lambda x: x.is_file(), sorted(starting_dir.glob(glob)))
    for item in files:
        choices.append(Choice(str(item.relative_to(starting_dir)), value=item))

    answer: Union[str, Path] = select(message, choices=choices).ask()
    if isinstance(answer, str):  # user chose Cancel
        return None
    if answer.is_file():
        return answer
    else:
        return questionary_file_select(message, answer, glob)
```

### LINKED ISSUE #73: Hiding cursor on select
Is it possible to hide the cursor for a select? If not, could it be possible?

image

Having the cursor there creates a visual distraction and I think also that it indicates that it is possible to type in text where the cursor is located.

### LINKED ISSUE #74: Provide an unsafe version for prompt()
This code: https://github.com/tmbo/questionary/blob/91facb2a78ab51855aae77d248f7b963c6cfc25a/questionary/prompt.py#L109-L113

It slurps user-cancellation undeniably.

It needs an `unsafe` argument that raises the proper exception.

### LINKED ISSUE #76: Validate_while_typing not working when select are included in questions
When running a prompt with a series of questions, I normally will use the `validate_while_typing=False` parameter to avoid validation errors imidiately when people start typing. This comes as you know from prompt toolkit itself, and are not implemented by Questionary.

But as soon as I introduce a select, things will fail like this:

```
questions = [
        {
            'type': 'text',
            'name': 'role_name',
            'message': 'Role name',
        },
        {
            'type': 'text',
            'name': 'role_description',
            'message': 'Role description',
        },
        {
            "type": "select",
            "name": "theme",
            "message": "What do you want to do?",
            "choices": [
                "Order a pizza",
                "Make a reservation",
                "Ask for opening hours",
                {"name": "Contact support", "disabled": "Unavailable at this time"},
                "Talk to the receptionist",
            ],
        }
    ]
    answers = prompt(questions, validate_while_typing=False)
```

Giving the error:

```
    answers = prompt(questions, validate_while_typing=False)
  File "python3.8/site-packages/questionary/prompt.py", line 95, in prompt
    question = create_question_func(**_kwargs)
  File "python3.8/site-packages/questionary/prompts/select.py", line 169, in select
    Application(layout=layout, key_bindings=bindings, style=merged_style, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'validate_while_typing'
```

It would be nice if turning validate_while_typing on and off also inside prompt. In fact, I think validate_while_typing should be False by default. It makes no sense to show an error to a user until he/she has been given a first chance to make it right.

### LINKED ISSUE #85: Print an output line with question styling
I tried going through the README, issues, examples and code, but couldn't really find a way.  
It's a bit weird to have a question and input that's nicely styled, and then as soon as I need to print something (like a confirmation, or just extra information), it falls out of style.  

One could argue it's not in the scope of this project, but I believe it's related enough (and should be easy enough to implement with what's already here).

tmbo:
- implemented unsafe version for prompt. fixes #74
- avoid passing unexpected arguments to application. fixes #76
- implemented print helper. fixes #85
- hiding cursor on select. fixes #73
- added path question type. fixes #44


tmbo:
@kiancross ready for review - tbh I was a bit lazy and combined all of the changes in this single PR. the changes are quite isolated, but in any case you could review by commits as well. let me know if that doesn't work

kiancross (approved):

examples/project_path.py
```diff
@@ -0,0 +1,8 @@
+import questionary
```
kiancross, line 1: I tested this example and a <kbd>backspace</kbd> did not redo the autocomplete. Not sure if this is a bug or intended.

kiancross (approved):
All looks good. Nice one!

Put a couple of minor things below, but think it looks good to go to go.

questionary/prompts/path.py
```diff
@@ -0,0 +1,129 @@
...
+            styled_display = completion.display[0]
+            # styled display is a formatted text (a tuple of the text and its style)
+            # second tuple entry is the text
+            if styled_display[1][-1] == "/":
+                # replace separator with the OS specific one
+                display_text = styled_display[1][:-1] + os.path.sep
+                # update the styledd display with the modified text
+                completion.display[0] = (styled_display[0], display_text)
+                # append the separator to the text as well - unclear why the normal
+                # path completer omits it from the text. this improves UX for the
```
layday, line 48: I believe the reason this is not appended automatically is that the slash is what triggers the initial completion.  If you start typing something, hit tab to complete it and then the slash is appended automatically for you, the completions for the next portion of the path won't appear - you'll have to hit backspace and re-insert the slash.  Hitting <tab> brings up the completion for the already-completed previous portion of the path.

tmbo: @layday `TAB` does bring up completion suggestions, at least for me?

the workflow I aimed for was: type, hit tab to open and select completion, hit enter to confirm completion, continue typing OR hit tab to complete the next segment

examples/project_path.py
tmbo: that's the same for my file auto completion (e.g. when using `cd`). a tab after that backspace will reopen the completion

questionary/prompts/path.py
```diff
@@ -0,0 +1,129 @@
...
+from prompt_toolkit.lexers import SimpleLexer
+from prompt_toolkit.shortcuts.prompt import CompleteStyle, PromptSession
+from prompt_toolkit.styles import Style, merge_styles
+
+from questionary.constants import DEFAULT_QUESTION_PREFIX, DEFAULT_STYLE
+from questionary.prompts.common import build_validator
+from questionary.question import Question
+
+
+class GreatUXPathCompleter(PathCompleter):
```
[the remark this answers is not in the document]
tmbo: not sure we really need the `min_input_len` but the `file_filter` is a good idea :+1:

examples/project_path.py
kiancross: Ok in that case it sounds like it's working as intended!

questionary/prompts/path.py
layday: It brings up the completion suggestions for the already-completed path:

completion

Notice how I had to insert a second `/` to bring up the suggestions for the next portion of the path.  Inserting a slash marks the end of the previous completion.

tmbo: you can also confirm the auto completion with enter instead of your second `/`

layday: Ah, I just assumed that enter would submit the answer regardless like prompt-toolkit would do.  It makes sense that way, thanks.

questionary/prompts/path.py
```diff
@@ -0,0 +1,141 @@
...
+
+def path(
+    message: str,
+    default: str = "",
+    qmark: str = DEFAULT_QUESTION_PREFIX,
+    validate: Any = None,
+    style: Optional[Style] = None,
+    only_directories: bool = False,
+    file_filter: Optional[Callable[[str], bool]] = None,
+    complete_style: CompleteStyle = CompleteStyle.MULTI_COLUMN,
```
layday, line 64: Perhaps you could also expose `expanduser` here so users will be able to complete from `~`?

layday:
Perhaps you could also expose `expanduser` in `path` so users will be able to complete from `~`?

questionary/prompts/path.py
tmbo: I've slightly changed the implementation, do you mind checking if it is more intuitive for you now?

layday: Sorry, I forgot to reply. It works great!

PR #87, opened 2020-11-02
Use arrow keys and shortcuts simultaneously

heeplr:
I needed shortcuts and arrow key item selection at the same time.

This small patch allows to do so by setting select(use_arrow_keys=True, use_shortcuts=True). Backwards compatible and not touching default behaviour (hopefully).


kiancross:
Looks good, thanks!

You just need to run `black .` to fix the styling issues, then all of the tests will pass.

A couple of tests would be good too :)

kiancross (approved):

PR #89, opened 2020-11-24
Fix double space between question and instructions in `confirm`

SamSchott:
At the moment, there are two spaces printed between a message and the instructions (Y/n) or answer in a `confirm dialog`.  This PR removes one of those spaces.

If the double spaces are intentional, feel free to close.

### DIFF
--- a/questionary/prompts/checkbox.py
+++ b/questionary/prompts/checkbox.py
@@ -91,7 +91,7 @@ def checkbox(
         if ic.is_answered:
             nbr_selected = len(ic.selected_options)
             if nbr_selected == 0:
-                tokens.append(("class:answer", " done"))
+                tokens.append(("class:answer", "done"))
             elif nbr_selected == 1:
                 if isinstance(ic.get_selected_values()[0].title, list):
                     ts = ic.get_selected_values()[0].title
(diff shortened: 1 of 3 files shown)


SamSchott:
On a side note: I noticed somewhat inconsistent spacing for several dialogs:

* Text: single space between message and answer / result.
* Path: single space
* Select: two spaces
* Checkbox: two spaces
* Confirm: two spaces

Personally, I prefer a single space. But either way, it would be nice to be consistent. It looks a bit odd otherwise.

kiancross:
Thanks for the PR. @tmbo do you have any views on 1 space vs 2 spaces? If not I suppose we should just pick one or the other for consistency.

tmbo:
mhm no not, really I think either one works - but all in on consistency 🚀

kiancross:
@SamSchott Do you want to change the others to single spaces too before I merge this in?

SamSchott:
Done.

kiancross (changes requested):
One more to remove here I think, then good to go.

https://github.com/tmbo/questionary/blob/146e320ebc1a5995fd2b7c6a0954d969396b3379/questionary/prompts/checkbox.py#L108

SamSchott:
I must have overlooked this one...

kiancross (approved):

PR #96, opened 2020-12-18
Change checkbox validation API to match other prompts

kiancross:
In #48 I added validation for checkboxes. I didn't realise until writing documentation that this API is not consistent with the function validation for other prompts (although having re-read the comments in the initial review, I think @tmbo did notice!). This PR makes validation for checkboxes consistent with other prompts (when using a function validator).


tmbo (approved):
looks great - thanks a lot for the fix :+1:

PR #97, opened 2020-12-18
Add documentation

### LINKED ISSUE #47: Documentation request: documentation for all different question choices
Had to dive through some examples to find stuff like 'when' and 'validate'. Would be nice with a sheet of available keys for each question.

kiancross:
There is a live version [here](https://questionary.readthedocs.io/en/latest/). At the moment it is pointed to my fork, once this is merged in I'll point it to this repo.

Tried to make the structure similar to other big Python projects so that it feels somewhat familiar.

@tmbo Do you have a ReadTheDocs account so I can make an admin?

Closes #47.


kiancross:
(sorry for all the spammy commits  - didn't realise they would all show in the PR 😂)

tmbo:
don't worry about the commits, we can squash them before merging. 

@kiancross yes I do have a rtd account, name is `tmbo`.

.readthedocs.yml
```diff
@@ -0,0 +1,17 @@
...
+formats:
+  - pdf
+
+sphinx:
+  fail_on_warning: true
+
+python:
+  version: 3.7
+  install:
+    - requirements: docs/source/requirements.txt
```
tmbo, line 15: should we use this workaround to avoid an additional requirements.txt? https://github.com/readthedocs/readthedocs.org/issues/4912#issuecomment-664002569

Makefile
```diff
@@ -39,3 +43,8 @@ test:
 
 types:
 	poetry run mypy questionary
+
+docs:
+	make -C docs html
+
+.PHONY: clean, install, formatter, lint, test, types, docs
```
tmbo, line 50: there is already a phony at the top, should we just add the additional ones there?

tmbo (commented):
really cool 🚀 ⭐  great work in porting this over to sphinx and separating the pages, cleaning up the code comments and making it run

I'll run it locally and do some edits right there, will be quicker than reviewing it in here

.readthedocs.yml
tmbo: looking into this right now, as I need to install docs locally anyways

kiancross: Yes agreed - this would be nicer.

Makefile
kiancross: I didn't notice that - I'll push a fix now!

tmbo (approved):
great work - let's merge it to get it deployed and I'll do my edits based on master 🔥

kiancross:
@tmbo Great! Can you add a webhook to the project so that the docs are built automatically? You should be able to add it automatically from this page: https://readthedocs.org/dashboard/questionary/integrations/.

kiancross:
Once a new version is released (i.e. there is a tag pointing to a commit which has the documentation), the `stable` build (which is the default when visiting `/`), will display the docs.

tmbo:
new docs are live 🎉 🚀 🥳  thanks a lot for making this happen :+1:

PR #98, opened 2020-12-18
Added auto_enter parameter to confirm

### LINKED ISSUE #37: Allow "confirmatory" enter to confirm dialog
When using the prompt-type confirm - it's really odd to me that pressing Y / n will immediately confirm the action and move forward. 

I think the current approach of auto-confirm will work fine for single, confirmatory questions - but can be a problem for workflows.

For that, i'd propose to have an option like "autoenter" - which should default to true (to keep existing behaviour) - however it'll allow myself to have the user press Y  / n - think about it - and then press enter explicitly.

The way current way is a problem as i find myself pressing "y" - "Enter" (obviously out of habit).

Since I'm building a workflow of multiple consecutive actions (all with defaults), i find myself constantly skipping / confirming the question right after the Confirm question - and i suspect that our users will face the same issue - if they're used to bash-like cli interfaces, which usually require a "n\r" - not just a "n".

Hope this makes sense

kiancross:
Closes #37

### DIFF
--- a/questionary/prompts/confirm.py
+++ b/questionary/prompts/confirm.py
@@ -99,13 +99,10 @@ def confirm(
 
     @bindings.add(Keys.ControlM, eager=True)
     def set_answer(event):
-
-        if status["answer"] is not None:
-            exit_with_result(event)
-
-        elif auto_enter:
+        if status["answer"] is None:
             status["answer"] = default
-            exit_with_result(event)
+
+        exit_with_result(event)
 
     @bindings.add(Keys.Any)
     def other(event):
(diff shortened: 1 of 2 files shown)


questionary/prompts/confirm.py
```diff
@@ -41,29 +42,36 @@ def confirm(
         style: A custom color and style for the question parts. You can
                configure colors as well as font types for different elements.
 
+        auto_enter: Does the user need to press the enter key to continue.
```
tmbo, line 45:
```suggestion
        auto_enter: If set to `False`, the user needs to press the 'enter' key to accept 
            their answer. If set to `True`, a valid input will be accepted without 
            the need to press 'Enter'.
```

tmbo (approved):
great implementation :+1: should be added to the docs as well once we have them 🙂

PR #101, opened 2020-12-20
Use direct image URLs in README so that they show on PyPi

kiancross:
The images in the README on PyPi are broken: https://pypi.org/project/questionary/

Using direct URLs should fix this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 * [Support](#support)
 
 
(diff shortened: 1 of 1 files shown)


kiancross:
No idea why coverage has decreased, as I haven't changed any code!

tmbo (approved):
good catch 👌

Issue #103, opened 2020-12-22
Configure SELECTED_POINTER


ElieDeloumeau:
Hello,

Is there a way to configure `SELECTED_POINTER`?

kiancross:
It's not possible at the moment, but it would be a nice enhancement.

PR #104, opened 2020-12-23
WIP: add pointer option

ElieDeloumeau:
related to #103 

(squash will be done when validation complete)

### DIFF
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -307,11 +307,12 @@ class InquirerControl(FormattedTextControl):
                 if self.pointer is not None:
                     tokens.append(("class:pointer", " {} ".format(self.pointer)))
                 else:
-                    tokens.append(("class:text", "   "))
+                    tokens.append(("class:text", " " * 3))
 
                 tokens.append(("[SetCursorPosition]", ""))
             else:
-                tokens.append(("class:text", "   "))
+                pointer_length = len(self.pointer) if self.pointer is not None else 1
+                tokens.append(("class:text", " " * (2 + pointer_length)))
 
             if isinstance(choice, Separator):
                 tokens.append(("class:separator", "{}".format(choice.title)))
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -180,7 +180,7 @@ class InquirerControl(FormattedTextControl):
     use_indicator: bool
     use_shortcuts: bool
     use_arrow_keys: bool
-    pointer: str
+    pointer: Optional[str]
     pointed_at: int
     is_answered: bool
 
@@ -188,7 +188,7 @@ class InquirerControl(FormattedTextControl):
         self,
         choices: Sequence[Union[str, Choice, Dict[str, Any]]],
         default: Optional[Union[str, Choice, Dict[str, Any]]] = None,
-        pointer: str = DEFAULT_SELECTED_POINTER,
+        pointer: Optional[str] = DEFAULT_SELECTED_POINTER,
         use_indicator: bool = True,
         use_shortcuts: bool = False,
         use_arrow_keys: bool = True,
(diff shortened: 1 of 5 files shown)


questionary/prompts/common.py
```diff
@@ -188,6 +188,7 @@ def __init__(
         self,
         choices: Sequence[Union[str, Choice, Dict[str, Any]]],
         default: Optional[Union[str, Choice, Dict[str, Any]]] = None,
+        pointer: str = DEFAULT_SELECTED_POINTER,
```
kiancross, line 191: Can you add a description in the docstring for this argument?

I think perhaps we could remove the `use_pointer` option and just set `pointer` to `None` to indicate that the pointer should be hidden.

kiancross (changes requested):
Thanks for the PR. It would also be nice to add this option to `checkbox`.

questionary/prompts/common.py
ElieDeloumeau: Sure it would be less redundant

ElieDeloumeau:
> Thanks for the PR. It would also be nice to add this option to `checkbox`.

Done 👍

questionary/prompts/common.py
```diff
@@ -180,27 +180,27 @@ class InquirerControl(FormattedTextControl):
     use_indicator: bool
     use_shortcuts: bool
     use_arrow_keys: bool
-    use_pointer: bool
+    pointer: str
```
kiancross, line 183:
```suggestion
    pointer: Optional[str]
```

questionary/prompts/common.py
```diff
@@ -180,27 +180,27 @@ class InquirerControl(FormattedTextControl):
...
     pointed_at: int
     is_answered: bool
 
     def __init__(
         self,
         choices: Sequence[Union[str, Choice, Dict[str, Any]]],
         default: Optional[Union[str, Choice, Dict[str, Any]]] = None,
+        pointer: str = DEFAULT_SELECTED_POINTER,
```
kiancross, line 191:
```suggestion
        pointer: Optional[str] = DEFAULT_SELECTED_POINTER,
```

questionary/prompts/rawselect.py
```diff
@@ -13,6 +13,7 @@ def rawselect(
     choices: Sequence[Union[str, Choice, Dict[str, Any]]],
     default: Optional[str] = None,
     qmark: str = DEFAULT_QUESTION_PREFIX,
+    pointer: str = DEFAULT_SELECTED_POINTER,
```
kiancross, line 16:
```suggestion
    pointer: Optional[str] = DEFAULT_SELECTED_POINTER,
```

questionary/prompts/select.py
```diff
@@ -19,11 +23,11 @@ def select(
     choices: Sequence[Union[str, Choice, Dict[str, Any]]],
     default: Optional[Union[str, Choice, Dict[str, Any]]] = None,
     qmark: str = DEFAULT_QUESTION_PREFIX,
+    pointer: str = DEFAULT_SELECTED_POINTER,
```
kiancross, line 26:
```suggestion
    pointer: Optional[str] = DEFAULT_SELECTED_POINTER,
```

kiancross (changes requested):
I think the `pointer` argument needs to be made optional so that `None` can be passed.

There will also be a problem if `pointer` is longer than a single character. I have tested a quick fix for this and I will push it to this PR if you are ok with that.

ElieDeloumeau:
> I think the `pointer` argument needs to be made optional so that `None` can be passed.
> 
> There will also be a problem if `pointer` is longer than a single character. I have tested a quick fix for this and I will push it to this PR if you are ok with that.

Of course you can 😃

ElieDeloumeau:
> There will also be a problem if `pointer` is longer than a single character. I have tested a quick fix for this and I will push it to this PR if you are ok with that.

Is this really a problem? The list is shifted but it doesn't matter?!

kiancross:
> Is this really a problem? The list is shifted but it doesn't matter?!

I think it is better if the choices remain aligned. Let me know if you are happy with the changes and then I will merge.

kiancross (approved):

PR #111, opened 2021-01-25
Use prompt-toolkit asyncio bug workaround for Windows

n8jhj:
BEFORE FIX:

```
(env) C:\...\questionary>python examples/readme.py
? What's your first name Nath
? What's your secret? ****
? Are you amazed? Yes

Unhandled exception in event loop:
  File "C:\Python38\lib\asyncio\proactor_events.py", line 768, in _loop_self_reading
    f.result()  # may raise
  File "C:\Python38\lib\asyncio\windows_events.py", line 808, in _poll
    value = callback(transferred, key, ov)
  File "C:\Python38\lib\asyncio\windows_events.py", line 457, in finish_recv
    raise ConnectionResetError(*exc.args)

Exception [WinError 995] The I/O operation has been aborted because of either a thread exit or an application request
Press ENTER to continue...
? What do you want to do? Make a reservation
? What do you want to do? Order a pizza

Unhandled exception in event loop:
  File "C:\Python38\lib\asyncio\proactor_events.py", line 768, in _loop_self_reading
    f.result()  # may raise
  File "C:\Python38\lib\asyncio\windows_events.py", line 808, in _poll
    value = callback(transferred, key, ov)
  File "C:\Python38\lib\asyncio\windows_events.py", line 457, in finish_recv
    raise ConnectionResetError(*exc.args)

Exception [WinError 995] The I/O operation has been aborted because of either a thread exit or an application request
Press ENTER to continue...
? Select toppings [bazz]
? Path to the projects version file questionary\__init__.py

(env) C:\...\questionary> 
```

AFTER FIX:

```
(env) C:\...\questionary>python examples/readme.py
? What's your first name Nath
? What's your secret? ****
? Are you amazed? Yes
? What do you want to do? Make a reservation
? What do you want to do? Order a pizza
? Select toppings [bazz]
? Path to the projects version file questionary\__init__.py

(env) C:\...\questionary>
```

Open issue (as of Jan. 24, 2021):
https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023

Thanks to Russell Hole for the workaround implementation example:
https://github.com/ctxis/cbrcli/pull/6/commits/9f8da07ff80cd3208f9623f5071ff9a92388ac2c

Code blackening left 17 files unchanged.

### DIFF
--- a/examples/readme.py
+++ b/examples/readme.py
@@ -1,3 +1,13 @@
+# Asyncio bug workaround
+# https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
+import asyncio
+import selectors
+
+selector = selectors.SelectSelector()
+loop = asyncio.SelectorEventLoop(selector)
+asyncio.set_event_loop(loop)
+# End workaround
+
 import questionary
 from examples import custom_style_dope
 
--- a/examples/readme.py
+++ b/examples/readme.py
@@ -1,13 +1,3 @@
-# Asyncio bug workaround
-# https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
-import asyncio
-import selectors
-
-selector = selectors.SelectSelector()
-loop = asyncio.SelectorEventLoop(selector)
-asyncio.set_event_loop(loop)
-# End workaround
-
 import questionary
 from examples import custom_style_dope
 
--- a/questionary/__init__.py
+++ b/questionary/__init__.py
@@ -2,6 +2,16 @@
 from prompt_toolkit.validation import Validator, ValidationError
 from prompt_toolkit.styles import Style
 
+# Asyncio bug workaround
+# https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
+import asyncio
+import selectors
+
+selector = selectors.SelectSelector()
+loop = asyncio.SelectorEventLoop(selector)
+asyncio.set_event_loop(loop)
+# End workaround
+
 import questionary.version
 from questionary.form import Form
 from questionary.form import form
--- a/questionary/__init__.py
+++ b/questionary/__init__.py
@@ -2,16 +2,6 @@
 from prompt_toolkit.validation import Validator, ValidationError
 from prompt_toolkit.styles import Style
 
-# Asyncio bug workaround
-# https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
-import asyncio
-import selectors
-
-selector = selectors.SelectSelector()
-loop = asyncio.SelectorEventLoop(selector)
-asyncio.set_event_loop(loop)
-# End workaround
-
 import questionary.version
 from questionary.form import Form
 from questionary.form import form


kiancross:
Thanks for the PR - are there any side effects to this? Does it need reverting when fixed upstream?

n8jhj:
@kiancross Sorry, I'm fairly new at using PRs. Could you please elaborate or point me to something I can read to get up to speed?

What is upstream? How do I make/test a fix upstream? What do you mean does it need reverting - are you asking whether this fix will not work further upstream? And how will I know whether there are any side effects?

kiancross:
Hi @n8jhj :)

You referenced prompt-toolkit/python-prompt-toolkit#1023, which in turn referenced [this](https://bugs.python.org/issue39010) bug. So it looks like there is a problem with Python (which is what I mean by 'upstream'). I was asking whether this fix needs reverting once the upstream issue (i.e. the Python issue) is fixed (i.e. is this fix temporary?).

The workaround you have given uses a `SelectorEventLoop` instead of a `ProactorEventLoop` - I am unfamiliar with the differences, so I was wondering if there would be any (negative) side effects to this change? E.g. could/would this cause problems for Unix users?

If you don't know the answers, don't worry - I can check myself when I have time :)

***

Does adding your fix before importing `questionary` also fix the issue?

n8jhj:
@kiancross Ah! Beautiful, thanks for clarifying. Looks like the necessary change has been merged to cpython master and backported to Python 3.8 and 3.9: https://github.com/python/cpython/pull/22066
Since questionary requires `python = "^3.6"`, I suppose you're stuck with this bugfix until you require at least 3.8.

Looks like Python made the `ProactorEventLoop` the default for Windows (yet, it apparently wasn't ready for production on Windows). Something to do with Windows only supporting sockets, not pipes: https://docs.python.org/3.6/library/asyncio-eventloops.html#available-event-loops
So I don't imagine there'd be any problems for Unix-based operating systems, since the `SelectorEventLoop` was designed for them.

<hr>

Yes, adding before questionary import still fixes it. In examples/readme.py:

```
# Asyncio bug workaround
# https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
import asyncio
import selectors

selector = selectors.SelectSelector()
loop = asyncio.SelectorEventLoop(selector)
asyncio.set_event_loop(loop)
# End workaround

import questionary
from examples import custom_style_dope

if __name__ == "__main__":
    ...
```

Confirmed that it breaks again when removing the workaround.

kiancross:
I believe that the fix has been released in [Python 3.8.7](https://docs.python.org/release/3.8.7/whatsnew/changelog.html) and [Python 3.9.1](https://docs.python.org/release/3.9.1/whatsnew/changelog.html#changelog) (<kbd>ctrl + f</kbd> for `bpo-39010`).

The change to make `ProactorEventLoop` the default apparently only happened in Python 3.8, which is why it has not been back-ported any further.

Can you check if this is fixed in the latest version of Python 3.8 or 3.9?

kiancross:
I believe this has been patched and backported in the latest Python versions so will close this PR :) If it's still an issue, just post back on this PR and we can reopen.

PR #116, opened 2021-02-11
[IMP] Pointer to colored hand

bosd:
On High-res display's the arrow pointer was not very visible. By changing to colored pointing finger. Readability is improved. Intuitively one will make a selection.

### DIFF
--- a/questionary/constants.py
+++ b/questionary/constants.py
@@ -16,7 +16,7 @@ NO_OR_YES = "(y/N)"
 INSTRUCTION_MULTILINE = "(Finish with 'Alt+Enter' or 'Esc, Enter')\n>"
 
 # Selection token used to indicate the selection cursor in a list
-DEFAULT_SELECTED_POINTER = "»"
+DEFAULT_SELECTED_POINTER = "👉"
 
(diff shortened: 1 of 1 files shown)


kiancross:
In the latest version (not released yet), the `pointer` can be customised: https://questionary.readthedocs.io/en/latest/pages/types.html#questionary.select

Does this work instead of chancing the default?

bosd:
Appearantly, tests do'nt like this proposal.
Don't know if it is my bad.

Was pointed to the proposal of having a custom pointer.
Would appreciate a default which is more readable. Does'nt have to be this emocon if that is not supported :)

kiancross:
I think it is better to leave the current pointer as is:

 - It can be changed using the [`pointer`](https://questionary.readthedocs.io/en/latest/pages/types.html#questionary.select) option.
 - Some terminals may not display emojis.
 - The current character (`»`) is ASCII and so has very good display support. I could not see an alternative suitable ASCII character to replace it with. Changing the default to a non-ASCII character may be a breaking change for some users.

If you can find a suitable replacement ASCII pointer character then feel free to open a new PR.

PR #125, opened 2021-05-03
Remove Poetry includes

layday:
These were being included in the wheel and ultimately installed
in `site-packages` by pip, i.e. Poetry's readme and license would reside
in `site-packages/{LICENSE,README.md}.  To limit includes to the sdist
you would have to use the belatedly documented table format:
https://github.com/python-poetry/poetry/pull/2268/files.
However, the readme and the license are included by Poetry in the sdist
by default, and py.typed is included both in the wheel and the sdist
by default.  The include directive here is redundant.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ classifiers=[
     "Topic :: Software Development :: Libraries",
 ]
 keywords = [ "cli", "ui", "inquirer", "questions", "prompt",]
-include = [ "LICENSE", "README.md", "questionary/py.typed",]
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


kiancross:
So just to double-check, the result of this change is only that the README and license are removed from the wheel and so not placed in `site-packages`?

layday:
Removed from the root of the wheel, yes.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

kiancross (approved):

PR #130, opened 2021-06-11
Added in `default="1"` for `Choice(title="bleh",value="1")`

WilliamStam:
the way "default" is handled currently isnt very intuitive. you need to pass a Choice object to it if you want it to use that choice's value

this PR lets you supply the default as just the "value" of the choice. 

as per 

https://github.com/tmbo/questionary/issues/124 and https://github.com/tmbo/questionary/issues/57 

(sorry not sure how to add tests for it, still noob)

PR #131, opened 2021-06-12
Add additional tests to `select` and `checkbox`

kiancross:
- Test when a `Choice` instance is passed as `default` or `initial_choice`.
 - Test when a `Choice` instance and string have the same value and `default` or `initial_choice` is set to the `Choice` instance.

Adding these to catch some of the breaking changes that would be introduced by #130.


tmbo (approved):
looks great. thanks a lot for the additional tests 🙂

Issue #132, opened 2021-06-20
`default` argument on `select` behaves differently when `Choice` is used


kiancross:
In this example, the title 'Bar' is highlighted:
```python3
questionary.select("Choose something", ["Foo", "Bar"], default="Bar").ask()
```

In this example, the title 'Bar' is highlighted:
```python3
c = questionary.Choice("Bar")
questionary.select("Choose something", ["Foo", c], default="Bar").ask()
```

In this example, the title 'Bar' is not highlighted:
```python3
c = questionary.Choice("Bar")
questionary.select("Choose something", ["Foo", c], default=c).ask()
```

The latter should either throw an error or behave the same as the first two examples.

PR #134, opened 2021-07-05
Fix highlighting in the questions of type select.

MAD-py:
issue #132 mentions an error in the highlighting of select type questions.

This error occurs when the default value is an object of type Choice because the condition being evaluated does not make a good comparison to know which option to highlight.

this is a quick, simple and functional fix that does not affect the rest of the operation.

### DIFF
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -254,8 +254,12 @@ class InquirerControl(FormattedTextControl):
             )
 
     def _is_selected(self, choice: Choice):
+        if isinstance(self.default, Choice):
+            compare_default = self.default == choice
+        else:
+            compare_default = self.default == choice.value
         return (
-            choice.checked or choice.value == self.default and self.default is not None
+            choice.checked or compare_default and self.default is not None
         ) and not choice.disabled
 
     def _assign_shortcut_keys(self):


kiancross (changes requested):
This looks great - thank you. Ideally, the broken example given in #132 would be added as a test case, however, I think this is a difficult example to unit test so it can be left out.

I think the only problem remaining is that `default` can also be a `dict`. At the moment, this isn't handled correctly either, I don't think.

https://github.com/tmbo/questionary/blob/2979fb8c6404f69112e8ebf918f9a2b29a310b45/questionary/prompts/common.py#L191

We could probably leave that for another PR, as I imagine the use case of passing a `dict` is far lower than passing a `Choice`.

kiancross (dismissed):

MAD-py:
I agree with what you say, I'm going to look at the other case this weekend and try to get a solution

But as you say I think that the best way to do it would be to create a new PR.

MAD-py:
I have a question about the use of default for the library that goes hand in hand with the solution I presented for this PR.

default if it should be sensitive to the data type in a strict way.

Example:

`
c = Choice("Bar")
select("Choose something", ["Foo", c], default="Bar").ask()
`

in this example two possible answers are sent, one of type str and the other of type choice, the default is of type str but with the same value as the answer of type choice.

The doubt that arises to me is if this error should be the expected one or if really one should be able to do this.

Output:
Captura de pantalla de 2021-07-10 11-53-51

kiancross:
The intention, as it stands, is for this to be an error. The `default` parameter is expected to identically match one of the `choices`. However, particularly for the case of passing a `dict`, this doesn't seem too nice.

I think there are a few aspects of the API that could be tidied up and perhaps this should be one of them.

kiancross (approved):

MAD-py:
I got it, so based on what you told me, in the next few days I will take a look at it and do some tests to see  how to improve the handling of the values set as dict.

I have just created an issue (#138) on the subject so we can continue talking about it.

thank you very much for clarifying my doubts and remember to close issue #132 that with the merge of this PR is ready.

PR #135, opened 2021-07-10
Fix incorrect dependabot file name

kiancross:
- Fixes incorrect dependabot file name.

 - Sets frequency to daily rather than weekly.


tmbo (approved):

tmbo:
lol explains why it didn'T check 😆

PR #157, opened 2021-08-22
Fix incorrect install target in Makefile

kiancross:
Without this, the `make docs` command does not work.


kiancross:
Will sort this out in another PR - don't think the requirements file should be there at all?

tmbo (approved):
yes good point :+1:

PR #158, opened 2021-08-22
Fix failing Coveralls CI

kiancross:
Coveralls tests usually fail when new PRs are submitted (it says the coverage has decreased).

I think this is because two coverage reports are being uploaded (one for Prompt Toolkit v2 and v3) without the `COVERALLS_PARALLEL` flag being set.

This PR uploads all coverage reports and indicates to Coveralls that they should be merged.

@tmbo I think the following check should also be added as required when this is merged: https://github.com/tmbo/questionary/pull/158/checks?check_run_id=3395118300

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -93,11 +93,24 @@ jobs:
       run: make test
 
     - name: Send Coverage Report 📊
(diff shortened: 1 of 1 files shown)


tmbo (approved):
lgtm 🚀

PR #161, opened 2021-08-23
Remove test preventing CI from completing on Windows with `prompt_toolkit` `2.*`

kiancross:
These checks should also be added to the list of required checks

### DIFF
--- a/tests/test_question.py
+++ b/tests/test_question.py
@@ -1,10 +1,13 @@
 import asyncio
+import pytest
+import platform
 
 from prompt_toolkit.input.defaults import create_pipe_input
 from prompt_toolkit.output import DummyOutput
 from pytest import fail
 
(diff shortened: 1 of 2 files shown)


tests/test_question.py
```diff
@@ -51,17 +52,19 @@ def test_skipping_of_skipping_of_questions():
         inp.close()
 
 
-def test_async_ask_question():
-    loop = asyncio.new_event_loop()
+if is_prompt_toolkit_3():
```
tmbo, line 55: I think there is a pytest primitive that does this kind of filtering. (I think using an annotation)

kiancross: @tmbo Thanks - I didn't know this. I've updated it.

tmbo (approved):

PR #172, opened 2021-10-06
Fix order of question-config-properties in unsafe_prompt()

### LINKED ISSUE #112: Support async/callables for choices argument
Currently, the `choices` argument must be a `Sequence`. The generation cost of this sequence can be high (e.g. IO bound, such an API call) which are called upon creation of the prompt and can create a latency to the start of script. In some cases, the prompt may be conditional and may never even be displayed. 

Consider this example:

```
def prompt_questions(difficult_decision):
    return questionary.unsafe_prompt(
        [
            {
                'type': 'confirm',
                'name': 'confirm',
                'message': f'Confirm decision {difficult_decision}',
                'when': lambda _: difficult_decision is not None,
                'default': False,
            },
            {
                'type': 'checkbox',
                'name': 'other_difficult_decisions',
                'message': 'Make your choices',
                'when': lambda x: not x.get('confirm'),
                'validate': lambda value: True if len(value) > 0 else 'You must select at least one decision',
                'choices': long_api_call_to_get_all_life_decisions(difficult_decision),
            },
        ]
    )
```

In this example, if the `confirm` stage returns `True`, `other_difficult_decisions` is never prompted, however, `long_api_call_to_get_all_life_decisions` was already called, creating a delay before `confirm` was prompted.

It would be great if instead a `callable` could be passed along and called at prompt time, in essence:

```
def prompt_questions(difficult_decision):
    return questionary.unsafe_prompt(
        [
            {
                ...
                'choices': functools.partial(long_api_call_to_get_all_life_decisions, difficult_decision),
            },
        ]
    )
```

With that said, first prize would be to support an asyncronous call as a choice, which is only yielded at prompt time.

HolgerDoerner:

When defining a prompt, the defined questions are processed in serial but the callable for the `Choices` is already called almost at the beginning. So `choices(answers)` is called always, even if it is unnecessary because the question should be skipped by condition.

Closes #112 

...

**How did you solve it?**

Move the execution of the `choices` callable AFTER the `when` -check. This way execution is prevented when question is skipped. This prevents side-effects and unnecessary delays when e.g. querying external resources.

...

**Checklist**

### DIFF
--- a/questionary/prompt.py
+++ b/questionary/prompt.py
@@ -137,10 +137,6 @@ def unsafe_prompt(
         if "name" not in question_config:
             raise PromptParameterException("name")
 
-        choices = question_config.get("choices")
-        if choices is not None and callable(choices):
-            question_config["choices"] = choices(answers)
-
         _kwargs = kwargs.copy()
         _kwargs.update(question_config)
 
@@ -166,6 +162,11 @@ def unsafe_prompt(
                 raise ValueError(
                     "'when' needs to be function that accepts a dict argument"
                 )
+
+        choices = question_config.get("choices")
+        if choices is not None and callable(choices):
+            question_config["choices"] = choices(answers)
+
         if _filter:
             # at least a little sanity check!
             if not callable(_filter):


tmbo:
thanks a lot for this improvement :+1:

questionary/prompt.py
```diff
@@ -166,6 +162,11 @@ def unsafe_prompt(
                 raise ValueError(
                     "'when' needs to be function that accepts a dict argument"
                 )
+
+        choices = question_config.get("choices")
+        if choices is not None and callable(choices):
+            question_config["choices"] = choices(answers)
```
tmbo, line 168: I think there might be an issue here - updating this in the `question_config` dict won't populate this change to the `_kwargs` variable with is used further down to create the question.

tmbo:
addressed in #208

PR #177, opened 2021-10-29
Minor/add start dirs for completion

### LINKED ISSUE #120: Allow start point of file path to be specified
Is there a way to specify a predefined start point when using the File Path type? For example, the user should always begin at 'C:\Users\x.xxxx\Documents' and navigates from there.

Thanks!

### LINKED ISSUE #127: Allow arbitrary paths to be displayed in file path input
I'm building an interface where the user is querying a remote file system. I love the File Path prompt and would like to use it, but here the files are not local.

It would be great if there was an option to provide an input with available file paths to use in the prompt.

NicDom:
- Allows the user to enter one or more starting directories for path completion.
- Validates that the given starting directories giving by `get_paths` by checking that these indeed are existing directories.

Closes #120. Possibly closes #127.

...

**How did you solve it?**
Setting an initial directory for path completion is already part of the [`prompt_toolkit`](https://github.com/prompt-toolkit/python-prompt-toolkit) [`PathCompleter`](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/5aea076692a304ec2bf8ad18fc59d4885cc462b1/prompt_toolkit/completion/filesystem.py#L13). It can be set via `get_paths`, which needs to be a callable returning a list of directories the completer then uses  for path completion, c.f. the documentation of `PathCompleter`:

> 
> ```python
> class PathCompleter(Completer):
>     """
>     Complete for Path variables.
> 
>     :param get_paths: Callable which returns a list of directories to look into
>                       when the user enters a relative path.
>    ...
>    """
> ```

Accordingly, referring to your example:
```python
get_paths = lambda: ["C:\Users\x.xxxx\Documents"]
```

Unfortunately, `get_paths` is not yet an argument of [`questionary.path.GreatUXPathCompleter`](https://github.com/tmbo/questionary/blob/c5b6066f8d43d9f90853c47b56e34ca227840479/questionary/prompts/path.py). Simply adding it to a call of [`questionary.path.path`](https://github.com/tmbo/questionary/blob/c5b6066f8d43d9f90853c47b56e34ca227840479/questionary/prompts/path.py), e.g. as
```python
path_ = path("Path to the projects version file", get_paths=lambda: ["C:\Users\x.xxxx\Documents"]).ask()
``` 
would lead to the error

```python
TypeError: __init__() got an unexpected keyword argument 'get_paths'
```

This is due to the reason that 'get_paths' is no known keyword and thus listed among `**kwargs`, which is given to [`prompt_toolkit.shortcuts.prompt.PromptSession`](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/5aea076692a304ec2bf8ad18fc59d4885cc462b1/prompt_toolkit/shortcuts/prompt.py#L228), which has no keyword 'get_paths'.
Thus, you need to add 'get_paths' to `questionary.path.path` and forward it to the `GreatUXPathCompleter`, i.e. add the line
```python
def path(
    message: str,
    default: str = "",
    validate: Any = None,
    only_directories: bool = False,
    start_dir: Optional[str] = None,
    get_paths: Optional[Callable[[], List[str]]] = None,   # add this line
    file_filter: Optional[Callable[[str], bool]] = None,
    complete_style: CompleteStyle = CompleteStyle.MULTI_COLUMN,
    **kwargs: Any,
) -> Question:
```

and make the change

```python
p = PromptSession(
        get_prompt_tokens,
        lexer=SimpleLexer("class:answer"),
        style=merged_style,
        completer=GreatUXPathCompleter(
            get_paths=get_paths, start_dir=start_dir, only_directories=only_directories, file_filter=file_filter, expanduser=True
        ),  # add 'get_paths' here
        validator=validator,
        complete_style=complete_style,
        key_bindings=bindings,
        **kwargs,
    )
```
in `questionary.path.path`.

Note, that no validation on 'get_paths' is performed, so adding an invalid path results in no path completion at all, without any warning, making it prone to errors. So to make it more stable, you may want to add some validation, e.g. in `GreatUXPathCompleter`:
```python
class GreatUXPathCompleter(PathCompleter):
    def __init__(
        self,
        only_directories: bool = False,
        get_paths: Optional[Callable[[], List[str]]] = None,
        file_filter: Optional[Callable[[str], bool]] = None,
        min_input_len: int = 0,
        expanduser: bool = False,
    ) -> None:
        # if get_paths is None, make it return the current working dir
        get_paths = get_paths or (lambda: ["."])
        # validation of get_paths
        for path in get_paths():
            if not os.path.isdir(path):
                raise (
                    ValueError(
                        "\n  'get_paths' must return only existing directories, but"
                        f" '{path}' is not. (at get_paths()[{get_paths().index(path)}])"
                    )
                )
        # call PathCompleter __init__
        super().__init__(
            only_directories=only_directories,
            get_paths=get_paths,
            file_filter=file_filter,
            min_input_len=min_input_len,
            expanduser=expanduser,
        )
```

...

**Checklist**

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,21 +24,21 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "20.3.0"
+version = "21.2.0"
 description = "Classes Without Boilerplate"
 category = "dev"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [package.extras]
-dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "furo", "sphinx", "pre-commit"]
-docs = ["furo", "sphinx", "zope.interface"]
-tests = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface"]
-tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six"]
+dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit"]
+docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
+tests = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface"]
+tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins"]
 
 [[package]]
 name = "babel"
-version = "2.9.0"
+version = "2.9.1"
 description = "Internationalization utilities"
 category = "main"
 optional = true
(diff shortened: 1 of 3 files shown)


NicDom:
Hi @kiancross,
thanks for your quick reply :)
Albeit #163 allows to fix #120, it does so in an not user-friendly way: One still has to write a `Completer` that solves the issue. With this PR, however, the issue is addressed immediately, as then one can simply pass a starting point to the prompt.
Nonetheless, I think its good to make the completer customizable. So I would suggest adding both features.

NicDom:
Added option to pass your own completer in order to include #163 into this **PR**.

kiancross:
Ok cool. I will have a closer look at this in a few weeks - the same time I have a look at your other PR. Thanks again for sending both of these :)

tmbo:
@NicDom thanks a lot for your contribution will take a look at it now - on first sight there are some changes to the python dependency lock file, I'll revert them as they don't seem intentional.

tmbo:
superseeded by #198. thanks a lot for your work @NicDom 🎉

PR #178, opened 2021-11-02
Feature/date

### LINKED ISSUE #121: Date input?
Would be cool if there was an intuitive way (kinda like the file path) to query a date/time.

NicDom:
- Fix #121
    - Add **Date and Time** prompts via `date` function
    - Add default validation and completion of text inputs for dates 
    - Allows user to easily use own or third-party-library date parsers for validation and completion
    - Own, customized validators and completers can also be used
- Add testsuite for new module
- Add new feature to documentation
- Add examples in documentation

...

**How did you solve it?**
Implemented `date` function, that returns a `Question` regarding a date/time input.
Prompting for dates/times is realized similar to path prompts:
Dates/times are to be entered via a text input, while the user consecutively offered different completions.

Two basic methods of completion and validation are implemented, to which I refer as the _simple_ and _parsing_ ones.

The _simple_ one depends on a given `date_format` consisting of  [`datetime` code formats](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes), e.g. `%Y-%m-%d`, and is expecting the user input to follow that format. It then offers completions according to the chosen format and input and validates the input using the format.
Note that currently only certain `date_formats` are supported. If a not supported one is entered, it raises a `ValueError`.

The _parsing_ one performs validation and completion using a given `parser`. If the `parser` is able to parse the input into a date, the input is considered to be valid and the parsed date-object is offered as a completion. 
The `date`-module implemented date-parser (`custom_date_parser`)  is by default used in the parsing method. Own parsers or third-part-libraries can be used.

Both completion and validation methods, the _simple_ and _parsing_ one, are then combined in the `date`-function to one
method, referred as _full_ on. This method is the default method used for completion and validation.

A typical would look like this:
```python3
>>> import questionary
>>> questionary.date("Type a date: ").ask()
# ? Type a date: 2021-01-01
# '2021-01-01'
```

or, for example using [`dateutil`](https://pypi.org/project/python-dateutil/) for 'parsing' validation and completion

```python3
>>> import questionary
>>> import dateutil.parser
>>> questionary.date("Type a date or time: ", parser=dateutil.parser.parse).ask()
# ? Type a date or time:  2021 jan 1
# '2021-01-01 00:00:00'
```
...

**Checklist**

### DIFF
--- a/questionary/prompts/date.py
+++ b/questionary/prompts/date.py
@@ -367,3 +367,64 @@ class FullDateCompleter(Completer):
             yield completion
         for completion in parsed_completions:
             yield completion
+
+
+class FullDateValidator(Validator):
+    """Validation of user input for :function: `date`.
+
+    Validates the user input for :func: `date` via calling validation of :class:
+    `ParsingDateValidator` (if ``parsing`` is a callable) or via calling ``validate`` of
+    :class: `SimpleDateValidator`.
+
+        Args:
+            date_format (str): Format determining the format that is to be used
+                for parsing :class: `datetime.date`.
+            parser (Callable[[str], AnyDate], optional): A callable that parses
+                a string into a :class: `datetime.date` or :class: `datetime.date`, e.g.
+                the ones from `dateparser`_ or `dateutil`_. Defaults to None.
+
+        .. _dateparser: https://github.com/scrapinghub/dateparser
+        .. _dateutil: https://github.com/dateutil/dateutil
+    """
+
+    def __init__(
+        self,
+        date_format: Optional[str] = ISO8601,
+        parser: Optional[Callable[[str], AnyDate]] = None,
+    ) -> None:
+        """__init__ of :class: `FullDateValidator`.
+
+        Validates the user input for :func: `date` via calling validation of :class:
+        `ParsingDateValidator` (if ``parser`` is a callable) or via calling
+        ``validate`` of :class: `SimpleDateValidator`.
+
+        Args:
+            date_format (str): Format determining the format that is to be used
+                for parsing :class: `datetime.date`. Defaults to ``ISO8601``.
+            parser (Callable[[str], AnyDate], optional): A callable that parses
+                a string into a :class: `datetime.date` or :class: `datetime.date`, e.g.
+                the ones from `dateparser`_ or `dateutil`_. Defaults to None.
+
+        .. _dateparser: https://github.com/scrapinghub/dateparser
+        .. _dateutil: https://github.com/dateutil/dateutil
+        """
+        self.date_format = date_format
+        self.parser = parser
+        self.simple_validator = SimpleDateValidator(date_format=date_format)
+        self.parsing_validator = ParsingDateValidator(parser=parser)
+
+    def validate(self, document: Document) -> None:
+        """Validaton of user input for :function: `date`.
+
+        Validates the user input for :func: `date` via calling validation of :class:
+        `ParsingDateValidator` (if ``parsing`` is a callable) or via calling
+        ``validate`` of :class: `SimpleDateValidator`.
+
+        Args:
+            document (Document): The :class: `prompt_toolkit.document.Document` created
+                of the user input.
+        """
+        if self.parser is None:
+            self.simple_validator.validate(document)
+        else:
+            self.parsing_validator.validate(document)
--- a/questionary/prompts/date.py
+++ b/questionary/prompts/date.py
@@ -557,7 +557,8 @@ def date(
             right of the prompt. Defaults to True.
         parser (Callable[[str], AnyDate], optional): A callable that parses
                 a string into a :class: `datetime.date` or :class: `datetime.date`, e.g.
-                the ones from `dateparser`_ or `dateutil`_. Defaults to None.
+                the ones from `dateparser`_ or `dateutil`_. If None, set to
+                :function: `custom_date_parser`.Defaults to None.
         return_date_object (bool): If True, a parsed date object is returned. Else, string
             the text input is returned. Defaults to True.
         complete_style (CompleteStyle): How autocomplete menu would be shown, it could
@@ -579,6 +580,8 @@ def date(
     """
     # delimeter used to separate year, month and day
     delimeter: str = date_format[-3]
+
+    parser = parser or custom_date_parser
 
     merged_style = merge_styles([DEFAULT_STYLE, style])
 
--- a/questionary/prompts/date.py
+++ b/questionary/prompts/date.py
@@ -131,7 +131,8 @@ def custom_date_parser(input: str) -> Optional[datetime.date]:
         "%d",
         "%H",
         "%M",
-        "%S"
+        "%S",
+        "%f"
     ]
     date_formats = [
         "".join(_time_format_codes[0:i]) for i in range(len(_time_format_codes))
--- a/questionary/prompts/date.py
+++ b/questionary/prompts/date.py
@@ -89,7 +89,7 @@ MINUTE = [str(i) for i in range(60)]
 SECOND = [str(i) for i in range(60)]
 
 # dict used to determine correct order for completions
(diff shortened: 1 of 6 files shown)


kiancross:
Hi @NicDom. This looks excellent - thank you! I won't have time to review this until December, but having briefly tried out the new date input and had a quick scan of the code, there should not be many issues :)

tmbo:
@NicDom thanks a lot for your contribution - will give it a review 🔥 (sorry for the delay, things have been crazy at work)

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+        # find fitting completions in completion_list
+        for entry in completion_list:
+            if entry.startswith(user_input):
+                completion = entry if int(entry) >= 10 else "0" + entry
+                completion += self.delimeter if len(text_split) < 3 else ""
+                yield Completion(
+                    old_input + completion,
+                    start_position=-len(document.text),
+                    style="fg:ansigreen bold",
+                    selected_style="fg:white bg:ansired bold",
```
kiancross, line 234: Can these styles be made to use one of the classes (or add a new class) so that they can be customised.

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+        """
+        parse_order = self.format.split(self.delimeter)
+        return [PARSE_FORMAT_DICT.get(item) for item in parse_order]  # type: ignore[misc]
+
+    def get_completions(
+        self, document: Document, complete_event: CompleteEvent
+    ) -> Iterable[Completion]:
+        """Completion of text input following a given ``date_format``.
+
+        According what has allready been typed and the given ``date_format``
```
kiancross, line 201:
```suggestion
        According what has already been typed and the given ``date_format``
```

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+        _date = _try_date_format(date_format=date_format, text=relevant_input)
+        # if parsing succeeded return the result
+        if _date is not None:
+            return _date
+    # no parsing succeeded so return None
+    return None
+
+
+###################################
+#  SIMPLE COMLETION AND VALIDATION
```
kiancross, line 160:
```suggestion
#  SIMPLE COMPLETION AND VALIDATION
```

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+        .. _datetime_parser: :function: `datetime.datetime.strptime`
+        """
+        try:
+            datetime.datetime.strptime(document.text, self.format)
+        except Exception:
+            raise (ValidationError(message="Invalid date."))
+
+
+###################################
+#  PARSING COMLETION AND VALIDATION
```
kiancross, line 282:
```suggestion
#  PARSING COMPLETION AND VALIDATION
```

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+            parsed_date = self.parser(document.text)
+        except Exception:
+            parsed_date = None
+        if self.parser is not None:
+            if parsed_date is None:
+                raise (ValidationError(message="Can not parse input to date object."))
+
+
+###################################
+#  FULL COMLETION AND VALIDATION
```
kiancross, line 388:
```suggestion
#  FULL COMPLETION AND VALIDATION
```

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+
+    parser = parser or custom_date_parser
+    if no_extra_parser:
+        parser = None
+
+    merged_style = merge_styles([DEFAULT_STYLE, style])
+
+    if print_date_format:
+        rprompt = to_formatted_text(
+            f"Date format: {date_format}", style="bg: ansigreen bold"
```
kiancross, line 604: Can these styles use a class (or a preexisting class) so that they can be customised.

kiancross (changes requested):
Thanks again for this PR. Some comments below.

- Unless they are needed for this PR, can you remove the changes to `poetry.lock`.
- The `date` function needs adding to the top-level `__init__.py` so that it can be accessed from the API.

NicDom:
Hi @tmbo  and @kiancross,

thanks for reviewing the PR. I will work on your suggested changes as soon as I can.

tmbo:
Thanks a lot @NicDom how are things going, ready for another review? 🙂

NicDom:
Hi,

sorry, have been "offline" the last days. I am going to look in to your suggested changes today.

questionary/prompts/date.py
```diff
@@ -0,0 +1,674 @@
...
+    "%d-%m-%Y",
+    "%m-%d-%Y",
+    "%m/%d/%Y",
+    "%m.%d.%Y",
+]
+
+# lists of completions (currently only numbers)
+DAY = [str(i) for i in range(1, 32)]
+MONTH = [str(i) for i in range(1, 13)]
+YEAR = ["0" * (4 - len(str(i))) + str(i) for i in range(9999)]
```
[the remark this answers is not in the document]
NicDom: Hi @kiancross,

yeah you are right. I'll guess the easiest way is to make the completer check, if a completion may lead to a valid date or not before yielding it.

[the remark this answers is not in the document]
NicDom: Hi,

ok, I added a validation of the completions before yielding it via adding the following method to `SimpleDateCompleter`:

```python
def _valid_completion(self, completion: str) -> bool:
        """Checks if a completion is a valid date.

        Used in :function: `get_completion` to reduce all options for completions
        to only those that lead to a valid date.
        Similar to :function: `custom_date_parser`, but does not ignore delimeters.

        Args:
            completion (str): The completion-option that is to be checked, if it
            leads to an valid date.
        """
        format_as_list = self.format.split(self.delimeter)
        date_formats = [
            self.delimeter.join(format_as_list[0:i])
            for i in range(1, len(format_as_list) + 1)
        ]

        # add delimeter to all but the last entry
        for i in range(len(date_formats) - 1):
            date_formats[i] += self.delimeter

        def _try_date_format(
            date_format: str, text: str
        ) -> Optional[datetime.datetime]:
            """Tries to parse ``text`to :class: `datetime.datetime`."""
            try:
                return datetime.datetime.strptime(text, date_format)
            except Exception:
                return None

        # try parsing for the several date_formats
        for date_format in date_formats:
            _date = _try_date_format(date_format=date_format, text=completion)
            # if parsing succeeded return True
            if _date is not None:
                return True
        # no parsing succeeded so return False
        return False
```
`_valid_completion` is then called before yielding the completion:
```python
# if completion leads to a valid date, yield this completion
if self._valid_completion(full_output):
    yield Completion(
        full_output,
        start_position=-len(document.text),
        style="fg:ansigreen bold",
        selected_style="fg:white bg:ansired bold",
        display=completion,
    )
```
I'll now use/introduce style classes for `style` and `selected_style` to make them customizable.

questionary/prompts/date.py
NicDom:
```suggestion
if print_date_format:
        rprompt = to_formatted_text(
            f"Date format: {date_format}", style="class:information"
        )
```

questionary/prompts/date.py
NicDom:
```suggestion
style="class:options",
selected_style="class:selected",
```
where
```python
("options", "fg: ansigreen bold"),  # options given by autocomplete
```
Moreover, I introduced a default style for `selected` in `date`, which differs from the one in `DEFAULT_STYLE`:

```python
# define a default style for selected completions. Gets overwritten if another
# format for selected is given by `style`
default_date_styles = Style([("selected", "fg: white bg: ansired bold")])

merged_style = merge_styles([DEFAULT_STYLE, default_date_styles, style])
```

NicDom:
Hi,
Have you had time to take a look at the changes? 🙂

NicDom:
Hey,
how are things going? How do you feel about the changes?

kiancross:
Hi @NicDom - I can take a look on the weekend.

PR #186, opened 2021-11-27
Question skip fix

amnuts:
Two situations I tried to use today were having questions set up in a `form` where some of those questions could be skipped, and setting up my own `KeyboardInterrupt` and so needing to use `unsafe_ask`, again with questions that could be skipped.

In both of those instances, the questions would not skip.

**How did you solve it?**
In the `questions.py` the check for skipping:
```python
        if self.should_skip_question:
            return self.default
```
was in the `ask` and `ask_async` functions.  I moved this check to the `unsafe_ask` and `unsafe_ask_async` functions instead.

This has the effect of fixing the issue when trying to implement a `KeyboardInterrupt` handler (so needing to use the `unsafe_*` functions), but also allowing skipping in the `form` because that directly calls `unsafe_ask`.  It also allows `ask` and `ask_async` to continue to work as they also just proxy to the unsafe version of the functions.

I added another couple tests to `test_question.py` for the `unsafe` version of the skip question tests, and I added another couple to the `test_form.py` file for testing of skipping questions.

image

**Checklist**

### DIFF
--- a/questionary/question.py
+++ b/questionary/question.py
@@ -38,9 +38,6 @@ class Question:
             `Any`: The answer from the question.
         """
 
-        if self.should_skip_question:
-            return self.default
-
         try:
             sys.stdout.flush()
             return await self.unsafe_ask_async(patch_stdout)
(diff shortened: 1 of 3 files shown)


tmbo (approved):
thanks a lot for this improvement, this makes a lot of sense :+1:

PR #191, opened 2021-12-19
fix(types): declare support for single questions

Yajo:

Without this patch, when calling `prompt` or `unsafe_prompt` with a `dict` on `questions`, mypy fails with:

    Argument 1 to "unsafe_prompt" has incompatible type "Dict[str, Any]"; expected "Iterable[Mapping[str, Any]]"

**How did you solve it?**

As dict is a valid supported input, I added it to the typing information, joined within an `Union` type.

**Checklist**


tmbo (approved):
thanks a lot for the improvement ⭐

PR #221, opened 2022-04-20
Add `develop` Makefile target and update contributor's guide

beliaev-maksim:

Clarify contributors guide and help new contributors to join easier

**How did you solve it?**

update make file and instructions

now poetry and pre-commit are configured and installed by `make develop`. Then, every git commit automatic lint check will run ensuring code quality

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ help:
 	@echo "make"
 	@echo "    clean"
 	@echo "        Remove Python/build artifacts."
-	@echo "    devinstall"
+	@echo "    develop"
 	@echo "        Configure development environment for questionary."
 	@echo "    install"
 	@echo "        Install questionary."
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ help:
 	@echo "make"
 	@echo "    clean"
 	@echo "        Remove Python/build artifacts."
+	@echo "    devinstall"
+	@echo "        Configure development environment for questionary."
 	@echo "    install"
 	@echo "        Install questionary."
 	@echo "    lint"
@@ -31,6 +33,11 @@ clean:
 
 install:
 	poetry install --extras "docs"
+
+devinstall:
+	pip install poetry
+	$(MAKE) install
+	pre-commit install
 
 lint:
 	poetry run pre-commit run -a
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean install lint test types docs livedocs
+.PHONY: clean install develop lint test types docs livedocs
 
 JOBS ?= 1
 
@@ -34,7 +34,7 @@ clean:
 install:
 	poetry install --extras "docs"
 
-devinstall:
+develop:
 	pip install poetry
 	$(MAKE) install
 	pre-commit install
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,7 @@ clean:
 install:
 	poetry install --extras "docs"
 
-develop:
-	pip install poetry
-	$(MAKE) install
+develop: install
 	poetry run pre-commit install
 
 lint:
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ install:
 develop:
 	pip install poetry
 	$(MAKE) install
-	pre-commit install
+	poetry run pre-commit install
 
 lint:
 	poetry run pre-commit run -a
--- a/docs/pages/contributors.rst
+++ b/docs/pages/contributors.rst
@@ -14,26 +14,40 @@ counts, so do not hesitate!
 2. Fork `the repository <https://github.com/tmbo/questionary>`_ on GitHub to
    start making your changes.
 
-3. Write some tests that show the bug is fixed or that the feature works as
-   expected.
+3. Configure development environment.
 
-4. Ensure your code passes the style checks by running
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -32,6 +34,11 @@ clean:
 install:
 	poetry install --extras "docs"
 
+develop:
+	pip install poetry
```
kiancross, line 38: I don't think this line should be included, as there are [many ways to install](https://python-poetry.org/docs/) poetry.

Makefile
```diff
@@ -32,6 +34,11 @@ clean:
...
+	$(MAKE) install
+	pre-commit install
```
kiancross, line 40: Should be called using poetry.

docs/pages/contributors.rst
```diff
@@ -14,28 +14,42 @@ counts, so do not hesitate!
...
    start making your changes.
 
-3. Write some tests that show the bug is fixed or that the feature works as
+3. Configure development environment.
+
+  a. create python virtual environment, eg
+
+  .. code-block:: console
+
+    virtualenv .env && source .env/bin/activate
```
kiancross, line 23: This should not be necessary when using poetry.

docs/pages/contributors.rst
```diff
@@ -14,28 +14,42 @@ counts, so do not hesitate!
...
+4. Write some tests that show the bug is fixed or that the feature works as
    expected.
 
-4. Ensure your code passes the style checks by running
+5. Ensure your code passes the code quality checks by running
 
   .. code-block:: console
 
-    $ black questionary
+    make lint
```
kiancross, line 38:
```suggestion
    $ make lint
```

docs/pages/contributors.rst
```diff
@@ -14,28 +14,42 @@ counts, so do not hesitate!
...
 
-5. Check all of the unit tests pass by running
+6. Check all of the unit tests pass by running
 
   .. code-block:: console
 
-    $ pytest --pycodestyle --cov questionary -v
+    make test
```
kiancross, line 44:
```suggestion
    $ make test
```

docs/pages/contributors.rst
```diff
@@ -14,28 +14,42 @@ counts, so do not hesitate!
...
 
-6. Check the type checks pass by running
+7. Check the type checks pass by running
 
   .. code-block:: console
 
-    $ mypy questionary
+    make types
```
kiancross, line 50:
```suggestion
    $ make types
```

kiancross (changes requested):
Thanks for the PR. Some comments below.

Makefile
beliaev-maksim: ideally, `Makefile` should make project ready to run. So far, installing `poetry` via pip looks a single cross platform solution.

docs/pages/contributors.rst
beliaev-maksim: what is the reason behind it?

docs/pages/contributors.rst
beliaev-maksim: correct, will remove

Makefile
kiancross: The Poetry developers [don't recommend](https://python-poetry.org/docs/#alternative-installation-methods-not-recommended) installing using pip. However, the contributor's guide could be updated to include a link to the Poetry installation page.

docs/pages/contributors.rst
kiancross: It is used to indicate that the command should be written in a shell. The `$` is [rendered differently](https://questionary.readthedocs.io/en/stable/pages/contributors.html) in the documentation to make this clear.

beliaev-maksim: Yes, that is true. However, this applies only to sphinx. Unfortunately, githib (where lots of developers can look for contribution docs) doesn't render $ sign, which becomes inconvenient to copy the command

Makefile
beliaev-maksim: Let me update the docs then

beliaev-maksim:
@kiancross 
I update the docs and make file.

I still hesitant to add `$` in commands due to GH rendering, eg https://github.com/tmbo/questionary/blob/master/docs/pages/contributors.rst#steps-for-submitting-code

as you see copy button will not be helpful here:
image

docs/pages/contributors.rst
```diff
@@ -14,28 +14,37 @@ counts, so do not hesitate!
 2. Fork `the repository <https://github.com/tmbo/questionary>`_ on GitHub to
    start making your changes.
 
-3. Write some tests that show the bug is fixed or that the feature works as
+3. Install Poetry using one of the described methods
+   in `official documentation <https://python-poetry.org/docs/#installation>`_
```
kiancross, line 18:
```suggestion
3. `Install Poetry <https://python-poetry.org/docs/#installation>`_.
```

kiancross (changes requested):

docs/pages/contributors.rst
kiancross: I prefer keeping the dollar sign because, by convention, it indicates that the command should be run in the shell.

I agree that the way the GitHub copy/paste works is a bit annoying, however, the documentation is intended to be viewed on the documentation website, rather than GitHub.

PR #236, opened 2022-06-17
Add missing Python version to CI matrix

kiancross:

Fixes a problem introduced in #222.

Some custom entries in the CI matrix do not have an explicit Python version, which does not appear to be supported in later versions of the `setup-python` action.

This should allow #233 to be merged.

**How did you solve it?**

I added the Python version to all runs of the CI workflow.

**Checklist**


tmbo (approved):
sounds good :+1:

PR #244, opened 2022-08-08
Add a `print` question type

### LINKED ISSUE #207: Add print question type to dictionary style prompt
### Describe the problem

With version `1.10.0` when I want to use both prompt with dictionary and `print` question type, I need to call several prompts and to merge answers:

```python
from pprint import pprint
from questionary import Separator, prompt
from questionary import print as qprint


def ask_dictstyle(**kwargs):
    questions = [
        {
            "type": "text",
            "name": "first",
            "message": "Your first message:",
        },
        {
            "type": "text",
            "name": "second",
            "message": "Your second message",
        },
    ]

    answers = prompt(questions)
    qprint("Oh! I need to give somme fancy explanations about what’s next!🦄", style="bold italic fg:darkred")
    questions = [
        {
            "type": "select",
            "name": "third",
            "message": "Select item",
            "choices": ["item1", "item2", Separator(), "other"],
        },
    ]
    answers.update(prompt(questions))
    return answers


if __name__ == "__main__":
    pprint(ask_dictstyle())

```

### Describe the solution

Adding `print` type to prompt dictionary would resolve the issue:

```python
from pprint import pprint
from questionary import Separator, prompt


def ask_dictstyle(**kwargs):
    questions = [
        {
            "type": "text",
            "name": "first",
            "message": "Your first message:",
        },
        {
            "type": "text",
            "name": "second",
            "message": "Your second message",
        },
        {
            "type": "print",
            "name": "help", #  Do I need a name?
            "message": "Oh! I need to give somme fancy explanations about what’s next!🦄",
            "style": "bold italic fg:darkred",
        },
        {
            "type": "select",
            "name": "third",
            "message": "Select item",
            "choices": ["item1", "item2", Separator(), "other"],
        },
    ]
    return prompt(questions)


if __name__ == "__main__":
    pprint(ask_dictstyle())
```

### Alternatives considered

_No response_

RhetTbull:
Adds `print` type to dictionary style prompts. Closes #207.

...

**How did you solve it?**
Modified `prompt()` to handle a `print` type that prints a message (with/without optional style).  The `name` field is optional when using `print`. `print` does not return a value and does not add an entry to the returned dict from `prompt()`; it only outputs text.  `print` does work with the `when` keyword so the message can be conditionally printed.

...

**Checklist**

- [X] I have read the [Contributor's Guide](https://questionary.readthedocs.io/en/stable/pages/contributors.html#steps-for-submitting-code).
- [X] I will check that all automated PR checks pass before the PR gets reviewed.

Note: several tests in `tests/prompts/test_select.py` failed with "too many open files".  However, these same tests failed *before* the PR changes.  I'm using a Mac running Catalina 10.15.7. All other tests (including the tests added for this PR) passed.

```
tests/prompts/test_select.py::test_jk_and_shortcut_conflict_avoided_by_disabling_ij_keys FAILED                                                 [ 94%]
tests/prompts/test_select.py::test_select_shortcuts FAILED                                                                                      [ 95%]
tests/prompts/test_select.py::test_select_no_arrow_keys FAILED                                                                                  [ 95%]
tests/prompts/test_select.py::test_select_no_shortcuts FAILED                                                                                   [ 96%]
tests/prompts/test_select.py::test_select_default_has_arrow_keys FAILED
```

mypy checks passed:

```
mypy questionary
Success: no issues found in 17 source files
```

The `pycodestyle` test required by contributor's guide had many failures (mostly long lines) that were *already* present in the source code -- not changed in this PR.  I fixed the changes in the PR to ensure they introduced no new pycodestyle errors. I did add `# noqa` to two lines in `prompt.py` that were 80 and 85 lines long respectively because changing the code would have been much less clean (would require an if/then block instead of an inline which was easier to read).

Note: running black on the `tests/` directory would have changed a bunch of the existing import statements so I made sure my new tests were blackified but did not run black on the existing code.

### DIFF
--- a/questionary/prompt.py
+++ b/questionary/prompt.py
@@ -11,6 +11,7 @@ from questionary import utils
 from questionary.constants import DEFAULT_KBI_MESSAGE
 from questionary.prompts import AVAILABLE_PROMPTS
 from questionary.prompts import prompt_by_name
+from questionary.prompts.common import print_formatted_text
 
 
 class PromptParameterException(ValueError):
@@ -143,7 +144,8 @@ def unsafe_prompt(
         # import the question
         if "type" not in question_config:
             raise PromptParameterException("type")
-        if "name" not in question_config:
+        # every type except 'print' needs a name
+        if "name" not in question_config and question_config["type"] != "print":
             raise PromptParameterException("name")
 
         _kwargs = kwargs.copy()
@@ -151,7 +153,7 @@ def unsafe_prompt(
 
         _type = _kwargs.pop("type")
         _filter = _kwargs.pop("filter", None)
-        name = _kwargs.pop("name")
+        name = _kwargs.pop("name", None) if _type == "print" else _kwargs.pop("name")
         when = _kwargs.pop("when", None)
 
         if true_color:
@@ -171,6 +173,20 @@ def unsafe_prompt(
                 raise ValueError(
                     "'when' needs to be function that accepts a dict argument"
                 )
+
+        # handle 'print' type
+        if _type == "print":
+            try:
+                message = _kwargs.pop("message")
+            except KeyError as e:
+                raise PromptParameterException("message") from e
+
+            # questions can take 'input' arg but print_formatted_text does not (it only outputs)
+            # Need to remove 'input', if present, to avoid breakage when testing or using custom input
+            _kwargs.pop("input", None)
+
+            print_formatted_text(message, **_kwargs)
+            continue
 
         choices = question_config.get("choices")
         if choices is not None and callable(choices):
--- a/questionary/prompt.py
+++ b/questionary/prompt.py
@@ -186,6 +186,8 @@ def unsafe_prompt(
             _kwargs.pop("input", None)
 
             print_formatted_text(message, **_kwargs)
+            if name:
+                answers[name] = None
             continue
 
         choices = question_config.get("choices")
--- a/questionary/prompt.py
+++ b/questionary/prompt.py
@@ -145,7 +145,7 @@ def unsafe_prompt(
         if "type" not in question_config:
             raise PromptParameterException("type")
         # every type except 'print' needs a name
-        if "name" not in question_config and question_config["type"] != "print": # noqa
+        if "name" not in question_config and question_config["type"] != "print":
             raise PromptParameterException("name")
 
         _kwargs = kwargs.copy()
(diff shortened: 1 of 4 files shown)


kiancross:
Thanks for this PR.

> The pycodestyle test required by contributor's guide had many failures (mostly long lines) that were already present in the source code -- not changed in this PR.

I think this part of the contributor's guide needs to be updated. You should just be able to run `make develop` and then `make lint` to apply all formatting rules. The `# noqa` comments should not be necessary, as the aforementioned commands will sort out all line length issues.

questionary/prompt.py
```diff
@@ -172,6 +174,20 @@ def unsafe_prompt(
...
+                message = _kwargs.pop("message")
+            except KeyError as e:
+                raise PromptParameterException("message") from e
+
+            # questions can take 'input' arg but print_formatted_text does not
+            # Remove 'input', if present, to avoid breaking during tests
+            _kwargs.pop("input", None)
+
+            print_formatted_text(message, **_kwargs)
+            continue
```
kiancross, line 189: I wonder if `answers[name] = None` if `name` is provided (rather than being a non-existent key in the answers dictionary)?

kiancross (commented):
This all looks great - thanks. I've left a comment below.

tests/test_prompt.py
```diff
@@ -47,3 +47,22 @@ def test_invalid_question_type():
...
+                    "type": "print",
+                }
+            ]
+        )
+
+
+def test_print_no_name():
+    """'print' type doesn't require a name so it
+    should not throw PromptParameterException"""
+    assert prompt([{"type": "print", "message": "Hello World"}]) == {}
```
kiancross, line 68: This test is breaking on Windows. It should work if you import `from prompt_toolkit.output import DummyOutput` and then pass `output=DummyOutput()` to `prompt`. See below for an example.

https://github.com/tmbo/questionary/blob/6643fe006f66802d9e504640f26b9e1a0a9d1253/tests/utils.py#L75

questionary/prompt.py
RhetTbull: I thought about including the print name in the answers dictionary as either:

1) setting the name to the print message
2) setting the name to `None` as you suggested

However, I think this is unnecessary and maybe even counterintuitive.  For one, it gives a different result whether or not `name` is provided. I don't think `name` should be required for a simple print.  Second, the python `print()` function returns `None` and I can't think of a case where I'd want to assign the return value of `print()` (which will always be `None`) to a variable.

I'm happy to change this behavior if you want me to but as a user of `questionary` it's not something I'd ever use.

kiancross: I'm not too keen on just ignoring the `name` argument, which is why I suggested returning `None`. However, you could instead throw an error if it is passed, stating that `name` is not supported for the `print` type. What do you think?

RhetTbull: I think it's cleaner to return `None` than throw an error as `name` is required for every other question type and someone who's just following the pattern of `{'name': name_value, 'type': type_value, ...}` might be initially puzzled about the error.  I'll modify the code to return None for name and add test.

kiancross (approved):

kiancross:
Looks good to me - thanks for the PR!

PR #259, opened 2022-10-05
Add shortcuts to navigate Checkbox lists faster

solanav:
When offering a very long list of choices in the Checkbox question, traversing it is very slow.

**How did you solve it?**
I added four new bindings to the following keys:
- Home: Go to the first choice.
- End: Go to the last choice.
- NextPage: Advance 10 positions.
- LastPage: Go back 10 postions.

**Checklist**

### DIFF
--- a/questionary/prompts/checkbox.py
+++ b/questionary/prompts/checkbox.py
@@ -246,6 +246,30 @@ def checkbox(
 
         perform_validation(get_selected_values())
 
+    @bindings.add(Keys.End, eager=True)
+    def move_cursor_start(event):
+        ic.pointed_at = 0
+
+    @bindings.add(Keys.Home, eager=True)
+    def move_cursor_end(event):
+        ic.pointed_at = len(ic.choices) - 1
+
+    @bindings.add(Keys.PageDown, eager=True)
+    def move_cursor_down_fast(event):
+        for _ in range(10):
+            ic.select_next()
+
+        while not ic.is_selection_valid():
+            ic.select_next()
+
+    @bindings.add(Keys.PageUp, eager=True)
+    def move_cursor_up_fast(event):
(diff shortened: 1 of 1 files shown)


kiancross:
Thanks for the PR, this looks great. Would you be able to add some unit tests?

questionary/prompts/checkbox.py
```diff
@@ -246,6 +246,30 @@ def all(_event):
...
+    def move_cursor_start(event):
+        ic.pointed_at = 0
+
+    @bindings.add(Keys.End, eager=True)
+    def move_cursor_end(event):
+        ic.pointed_at = len(ic.choices) - 1
+
+    @bindings.add(Keys.PageDown, eager=True)
+    def move_cursor_down_fast(event):
+        for _ in range(10):
```
kiancross, line 259: Does everything work correctly if there are fewer than 10 items? For example, if there are 9 items, page down should go to the 9th. Could you write a test for this?

questionary/prompts/checkbox.py
```diff
@@ -246,6 +246,30 @@ def all(_event):
...
+            ic.select_next()
+
+        while not ic.is_selection_valid():
```
kiancross, line 262: What happens if you have a list with 1 valid item and 99 invalid items? In such a case, I think this might cause an infinite loop (unless `ic.select_next()` loops around to the start of the list)? Can you write a test to cover this.

kiancross (changes requested):
Sorry for the delay in reviewing this. I've left some questions/comments below.

baturayo:
Great work thanks lot! I hacked the paging in my application. For example, for my use case I need to have 20 items in one page. Can we make 10 items dynamic and configurable? @solanav

PR #270, opened 2023-04-15
Deprecate Python 3.6 & 3.7; get CI working again

BrianPugh:
# Change Log
* Dropped python 3.6 and python 3.7 support.
* Added python 3.10 and python 3.11 to all the relevant places
* Updated pre-commit hooks to latest versions
* Replaced deprecated isort flag `-sl` with current `--force-single-line-imports`.
* Change "extras" dependencies to modern poetry dependency groups.
    * Removed all sphinx-related doc dependencies from the main dependency group.
* Fixed a few minor typing issues with `merge_styles`
* Explicitly set the `prompt_toolkit` version to be `<=3.0.36` because `3.0.37` has introduced some event-loop-related bug.
* `make clean` will now also clean the docs.

# Original Post
**What is the problem that this PR addresses?**
* isort 5.10.1 has the issues described [here](https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co).

**How did you solve it?**
* Ran `pre-commit autoupdate` to update all plugins; fixing minor linting issues in changes of Black.

**Checklist**

- [x] I have read the [Contributor's Guide](https://questionary.readthedocs.io/en/stable/pages/contributors.html#steps-for-submitting-code).
- [x] I will check that all automated PR checks pass before the PR gets reviewed.

PR #271, opened 2023-04-15
Implement 'press any key to continue' prompt

BrianPugh:
* Previously, there was no convenient way of prompting "Press any key to continue...".
* A similar effect could be achieved using the `confirm` prompt, but it would display yes/no, which may confuse the user.
* Built ontop of #270 so that pre-commit can actually run.

**How did you solve it?**
* Implements a "Press any key to continue..." prompt, modeled after the `confirm` prompt.
* We can populate more documentation and stuff if we are happy with this feature.

**Example Use Case**

```
from questionary import press_any_key_to_continue

# Block until the user presses a key
press_any_key_to_continue().ask()  # Prints " Press any key to continue..."
```

**Checklist**

### DIFF
--- a/questionary/prompts/press_any_key_to_continue.py
+++ b/questionary/prompts/press_any_key_to_continue.py
@@ -1,0 +1,61 @@
+from typing import Optional
+
+from prompt_toolkit import PromptSession
+from prompt_toolkit.formatted_text import to_formatted_text
+from prompt_toolkit.key_binding import KeyBindings
+from prompt_toolkit.keys import Keys
+from prompt_toolkit.styles import Style
+from prompt_toolkit.styles import merge_styles
+
+from questionary.constants import DEFAULT_STYLE
+from questionary.question import Question
+
+
+def press_any_key_to_continue(
+    message: Optional[str] = None,
+    style: Optional[Style] = None,
+    **kwargs,
+):
+    """Wait until user presses any key to continue.
+
+    Example:
+        >>> import questionary
+        >>> questionary.press_any_key_to_continue().ask()
+         Press any key to continue...
+        None
+
+    Args:
+        message: Question text. Defaults to `"Press any key to continue..."`
+
+        style: A custom color and style for the question parts. You can
+               configure colors as well as font types for different elements.
+
+    Returns:
+        :class:`Question`: Question instance, ready to be prompted (using `.ask()`).
+    """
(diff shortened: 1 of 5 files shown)


kiancross:
Looks great, thank you! Can you add the question to the docs:

https://github.com/tmbo/questionary/blob/master/docs/pages/types.rst

BrianPugh:
@kiancross this has been rebased to master, and docs have been updated.

questionary/prompts/press_any_key_to_continue.py
```diff
@@ -0,0 +1,62 @@
...
+    style: Optional[Style] = None,
+    **kwargs: Any,
+):
+    """Wait until user presses any key to continue.
+
+    Example:
+        >>> import questionary
+        >>> questionary.press_any_key_to_continue().ask()
+         Press any key to continue...
+        ''
```
kiancross, line 26: Is this correct? Or should it be `None`?

kiancross (commented):
All looks fine! Quick question below. Also the code format CI is failing. Thanks again.

questionary/prompts/press_any_key_to_continue.py
BrianPugh: I think it might be related to the version of `prompt_toolkit`? I distinctly remember that when I initially programmed it, it returned `None`, but when I tried it again recently it returned an empty string `''`.

BrianPugh:
addressed! related (in another PR): we should add mypy to pre-commit.

questionary/prompts/press_any_key_to_continue.py
kiancross: I think it must be None, given that assert result is None is used in the test?

BrianPugh: This is with prompt_toolkit `3.0.36`.

Screenshot 2023-05-25 at 12 18 44 PM

yet, at the same time, the unit test passes. So it seems like the `feed_cli_with_input` does some sort of implicit empty-string-to-none conversion. In real applications, it returns an empty string, we can try and address the unit-test inconsistency in another PR.

kiancross: Ok sounds good!

kiancross (approved):
Thanks again!

PR #334, opened 2023-11-12
fix: disabled Choice binding shortcut_key

weiduhuo:

When defining a `Choice` type option with simultaneously setting the `shortcut_key` attribute to `str` type and setting the 
 `disabled` attribute to be not `None`, there is an unexpected ability that we can interact with this option using keyboard shortcuts. 

example:
```py
import questionary
from questionary import Choice

options = [
    Choice("One", disabled=True),
    Choice("Three", shortcut_key="3", disabled=True),
    Choice("Two"),
    Choice("Four"),
]
questionary.select("Title: ", options, use_shortcuts=True).ask()
```

**How did you solve it?**

I cancel the disabled choice binding the key.

**Checklist**

### DIFF
--- a/questionary/prompts/select.py
+++ b/questionary/prompts/select.py
@@ -200,7 +200,7 @@ def select(
                     "for movement are disabled. "
                     "This choice is not reachable.".format(c.title)
                 )
-            if isinstance(c, Separator) or c.shortcut_key is None:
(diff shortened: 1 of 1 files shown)


kiancross (approved):
Thanks, looks good to me. Can you force-push the commit to re-trigger the CI?

PR #340, opened 2023-11-28
fix: shortcut error in loop

weiduhuo:

When we use the `select` method within a loop, and the `Choice` parameter sequence is defined outside the loop, shortcut keys in alternating loops are presented in a doubled manner of the actual values.

example:
```py
import questionary
from questionary import Choice

options = [Choice("One"), Choice("Two"), Choice("Three")]
for _ in range(4):
    option = questionary.select("Title:", options, use_shortcuts=True).ask()
```

**How did you solve it?**

The root cause is that, during the initialization of the `InquirerControl.choices` in the `_init_choices` process, when a `Choice` object is instantiated with another `Choice` object as a parameter, it directly references and passes the return result. This causes the `shortcut_key` parameter state of the `Choice` object to be retained in the loop, leading to the aforementioned issue.
We can resolve this issue by modifying the `build` method of the `Choice` class. When accepting a Choice-type parameter, we can return the result using a shallow copy."

```py
import copy

if isinstance(c, Choice):
   return copy.copy(c)
```

**Checklist**

### DIFF
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -1,3 +1,4 @@
+import copy
 import inspect
 from typing import Any
 from typing import Callable
@@ -114,7 +115,7 @@ class Choice:
         """
 
         if isinstance(c, Choice):
-            return c
+            return copy.copy(c)
         elif isinstance(c, str):
             return Choice(c, c)
         else:
--- a/questionary/prompts/common.py
+++ b/questionary/prompts/common.py
@@ -279,7 +279,7 @@ class InquirerControl(FormattedTextControl):
 
     def _is_selected(self, choice: Choice):
         if isinstance(self.default, Choice):
-            compare_default = self.default == choice
+            compare_default = self.default.value == choice.value
         else:
             compare_default = self.default == choice.value
         return choice.checked or compare_default and self.default is not None


kiancross:
Thanks for spotting this bug and submitting a PR. From what I understand, the problem is as follows:

1. When a `Choice` is instantiated without a shortcut, `Choice.shortcut_key` is `None` and so `Choice.auto_shortcut` is set to `True`.
2. Choices passed to `InquirerControl` are untouched if they are an instance of a `Choice` (and so any changes made to the choice by `InquirerControl` are preserved across calls).
3. `_assign_shortcut_keys` iterates through `choices` and removes shortcuts which have already been assigned from the list.
4. Remaining shortcuts are then assigned to choices with `auto_shortcut` set to `True`.

The problem occurs on subsequent calls to (3) and (4).

I see a few possible solutions:

1. The one you have proposed.
2. Once `Choice.shortcut_key` is set, `Choice.auto_shortcut` should be set to `False`. So the following logic should be moved into a setter for `Choice.shortcut_key`:
https://github.com/tmbo/questionary/blob/1c66330922c2f0da799553b8e84211bf4d1ee3fb/questionary/prompts/common.py#L93-L102
3. Either adding `and not c.auto_shortcut` here: https://github.com/tmbo/questionary/blob/1c66330922c2f0da799553b8e84211bf4d1ee3fb/questionary/prompts/common.py#L292
or `c.shortcut_key is not None` here:
https://github.com/tmbo/questionary/blob/1c66330922c2f0da799553b8e84211bf4d1ee3fb/questionary/prompts/common.py#L306

What do you think?

PR #374, opened 2024-05-29
feat: add search functionality to select and checkbox prompt, based on #42

### LINKED ISSUE #33: Add prefix filter search in select control
As featured here (when selecting the AWS region)
https://asciinema.org/a/AyujEPdjcDmSPoOK26pTozCiO

@tmbo thanks for pointing me here from PyInquirer. Looks like the more modern/maintained alternatives (latest lib, tests, ...)

I'll try to use questionary soon and if it's all good, I'll contribute the filter that I have built that I find very practical on long lists like the one I'm dealing with.

mario-dg:
This commit is heavily inspired by [gbataille's](https://github.com/gbataille) [PR](https://github.com/tmbo/questionary/pull/42)

Continued on the work mentioned above to integrate a search functionality to the `select` and `checkbox` prompt.
Should fix #33 and #42.

**How did you solve it?**

Taken the code from the previous PR, modified a bit to fit the latest updates, corrected type checking and linting.

**Checklist**


kiancross:
@mario-dg Thanks for this PR. Please can you merge latest changes into this branch, then I will run the CI and review.

mario-dg:
@kiancross, done 😄

questionary/prompts/checkbox.py
```diff
@@ -179,8 +194,9 @@ def get_prompt_tokens() -> List[Tuple[str, str]]:
                         "class:instruction",
                         "(Use arrow keys to move, "
                         "<space> to select, "
```
kiancross, line 196: Are we happy to say that space can't be used in the search?

questionary/prompts/common.py
```diff
@@ -343,9 +347,19 @@ def _init_choices(
 
             self.choices.append(choice)
 
+    @property
+    def filtered_choices(self):
+        if not self.search_filter:
+            return self.choices
+        filtered = [
+            c for c in self.choices if self.search_filter.lower() in c.title.lower()
```
kiancross, line 355: I'm ok with this search mechanism, but it is not a prefix search like #42. Does this matter? Are there any downsides to searching for any occurrence of the substring? I suppose in the future, we can make the filter customisable, so it doesn't matter too much...

pyproject.toml
```diff
@@ -9,7 +9,7 @@ exclude = "((.eggs | .git | .pytype | .pytest_cache | build | dist))"
 
 [tool.poetry]
 name = "questionary"
-version = "2.0.1"
+version = "2.1.0"
```
kiancross, line 12:
```suggestion
version = "2.0.1"
```

The version will be bumped when we make a new release.

questionary/version.py
```diff
@@ -1 +1 @@
-__version__ = "2.0.1"
+__version__ = "2.1.0"
```
kiancross, line 1:
```suggestion
__version__ = "2.0.1"
```

The version will be bumped when we make a new release.

kiancross (changes requested):
This PR is fantastic - thank you. I have tested it on my machine, and it all appears to work as expected.

I have left a couple of comments below, just to check you are happy with all the design decisions, but I think it is almost ready to merge (once the version numbers are fixed).

Thanks again.

questionary/prompts/common.py
mario-dg: After using the prefix search for a few days it felt unnatural and not intuitive. I personally don't know any application that prefers a prefix search above a substring search. 

A fuzzy search would probably be the best, but add unnecessary complexity and/or dependencies to this library.

questionary/prompts/checkbox.py
mario-dg: We have to exclude it, since space is used to select an entry from the list. Otherwise we would have to come up with a rather complex solution to allow (de-)selection of an entry when the search is enabled.

kiancross: Makes sense

questionary/prompts/common.py
kiancross: Thanks for explaining

kiancross (approved):
Looks good to me. Thanks!

PR #411, opened 2024-11-26
allow Path object to be passed as default

Kobu:

More oftan than not, when we prompt the user for a path, we later construct a builtin `Path` object from the received string. Following this notion, it is only logical that such object can also be passed as `default` parameter to the `questionary.path()` function.

**How did you solve it?**

We expand the type of `default` parameter to `Path | str` while also keeping its default value to be an empty string. Then we change the reference to this parameter on line 242 and we cast it to string using `str(default)`. If `default` is of type `Path`, then we recieve the corresponding path as a string, if its type is `str` then nothing happens.

**Checklist**

### DIFF
--- a/questionary/prompts/path.py
+++ b/questionary/prompts/path.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 from typing import Any
 from typing import Callable
 from typing import Iterable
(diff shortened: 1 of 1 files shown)


kiancross (approved):
Thanks for the PR. Please could you update with the latest changes from `master` and then I will merge (this will also re-run the tests which haven't been run yet).

PR #421, opened 2025-01-20
 Add modifications to set the passwordinvisible

Celina1801:

...

**How did you solve it?**

...

**Checklist**

- [ ] I have read the [Contributor's Guide](https://questionary.readthedocs.io/en/stable/pages/contributors.html#steps-for-submitting-code).

### DIFF
--- a/questionary/prompts/password.py
+++ b/questionary/prompts/password.py
@@ -1,9 +1,10 @@
+import getpass
 from typing import Any
 from typing import Optional
 
 from questionary import Style
 from questionary.constants import DEFAULT_QUESTION_PREFIX
-from questionary.prompts import text
+from questionary.prompts.common import build_validator
 from questionary.question import Question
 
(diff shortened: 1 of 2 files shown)


questionary/prompts/password.py
```diff
@@ -55,7 +56,26 @@ def password(
     Returns:
         :class:`Question`: Question instance, ready to be prompted (using ``.ask()``).
     """
+    # Créer un validateur si nécessaire
+    validator = build_validator(validate)
 
-    return text.text(
-        message, default, validate, qmark, style, is_password=True, **kwargs
-    )
+    def ask():
```
tmbo, line 62: mhm I am not quite sure why this is needed, the `is_password=True` should already ensure that the password is not visible and replaced with `*`. Can you describe a bit more what the advantace advantage of getpass is?

PR #425, opened 2025-01-27
Add dite/time input with default value with a defined range

### LINKED ISSUE #121: Date input?

RafikLkr:

This PR addresses the need to verify the handling of default values and the validation of date input prompts. Specifically, it ensures that when the user presses "Enter" without providing a date, the default date value is correctly returned. Additionally, this PR introduces functionality to validate if the input date falls within a defined range, with a minimum and maximum date that can be specified.

Closes #121 (assuming this is the issue related to the functionality being tested)

**How did you solve it?**

I implemented a unit test for the date input prompt. The test simulates a scenario where the user presses "Enter" without providing a date, which should return the default value specified (in this case, "2023-01-01"). The test uses create_pipe_input to simulate the user input and DummyOutput to handle output, verifying that the expected default value is returned correctly.

**Checklist**

- [ x] I have read the [Contributor's Guide](https://questionary.readthedocs.io/en/stable/pages/contributors.html#steps-for-submitting-code).
- [ x] I will check that all automated PR checks pass before the PR gets reviewed.

### DIFF
--- a/questionary/__init__.py
+++ b/questionary/__init__.py
@@ -23,6 +23,7 @@ from questionary.prompts.press_any_key_to_continue import press_any_key_to_conti
 from questionary.prompts.rawselect import rawselect
 from questionary.prompts.select import select
 from questionary.prompts.text import text
+from questionary.prompts.date import date
 from questionary.question import Question
 
 __version__ = questionary.version.__version__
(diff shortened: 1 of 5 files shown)


tmbo (approved):
looks great, thanks for this additional feature

PR #426, opened 2025-01-27
Require typed answer for confirm

### LINKED ISSUE #312: Require typed answer for confirm question
### Describe the problem

When I use the "confirm" question, I can force the user to use the "Enter" key to answer the question but I cannot require him to type anything at all.
If the user decides to just spam "Enter", all defaults values will be used instead.

For some important confirmations like "Are you sure you want to delete that file ?" or something similar, I want to be sure the user types something to answer (y/n or Yes/no) and cannot press Enter only.

I can set the default value to None (or anything different from True/False) and detect that answer but I would need to re-throw the question again instead of locking the user in the question.

"validate" isn't supported in "confirms" also

Basically I would need for a "y/n" instead of a "Y/n" or "y/N".

### Describe the solution

Add a "mandatory" argument of sort to the "confirm" type to force the user to type something and not leave the answer empty.

### Alternatives considered

Using a custom text question or choice question but this basically feels like reinventing the wheel at that point.
The first solution I had was to re-throw the same question until it is answered correctly like I said before but that's not "pretty".

Celina1801:

Closes #312
The confirm prompt currently allows users to bypass input by pressing "Enter," which automatically selects the default value. This behavior is problematic for critical confirmations (e.g., deleting files), where explicit user input (e.g., "y" or "n") should be required to ensure intentionality.

**How did you solve it?**

I introduced a mandatory parameter to the confirm prompt. When mandatory=True, the prompt enforces explicit user input ("y" or "n") and rejects empty or invalid inputs. The implementation modifies the key bindings to re-prompt the user until a valid response is provided.

Unit tests were added to ensure that:

Empty inputs are rejected when mandatory=True.
Valid responses ("y," "n," "Y," "N") are accepted.
Invalid inputs re-prompt the user for confirmation.

**Checklist**

- [ ] I have read the [Contributor's Guide](https://questionary.readthedocs.io/en/stable/pages/contributors.html#steps-for-submitting-code).

### DIFF
--- a/questionary/prompts/confirm.py
+++ b/questionary/prompts/confirm.py
@@ -18,36 +18,25 @@ from questionary.styles import merge_styles_default
 
 def confirm(
     message: str,
-    default: bool = True,
+    default: Optional[bool] = True,
     qmark: str = DEFAULT_QUESTION_PREFIX,
     style: Optional[Style] = None,
     auto_enter: bool = True,
     instruction: Optional[str] = None,
+    mandatory: bool = True,
     **kwargs: Any,
 ) -> Question:
     """A yes or no question. The user can either confirm or deny.
 
     This question type can be used to prompt the user for a confirmation
     of a yes-or-no question. If the user just hits enter, the default
-    value will be returned.
-
-    Example:
-        >>> import questionary
-        >>> questionary.confirm("Are you amazed?").ask()
-        ? Are you amazed? Yes
-        True
-
-    .. image:: ../images/confirm.gif
-
-    This is just a really basic example, the prompt can be customised using the
-    parameters.
-
+    value will be returned unless mandatory=True.
 
     Args:
         message: Question text.
 
-        default: Default value will be returned if the user just hits
-                 enter.
+        default: Default value will be returned if the user just hits enter,
+                 unless mandatory=True.
 
         qmark: Question prefix displayed in front of the question.
                By default this is a ``?``.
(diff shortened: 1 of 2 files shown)


tests/prompts/test_confirm.py
```diff
@@ -5,23 +5,8 @@
 from tests.utils import feed_cli_with_input
 
 
-def test_confirm_enter_default_yes():
```
tmbo, line 8: Any reason you removed these tests? The default behaviour should stay the same

tmbo:
I think this is a good addition, do you want to follow up on the comment I made on the removed tests? If you re-add them I think we are good to merge @Celina1801

questionary/prompts/confirm.py
```diff
@@ -18,36 +18,25 @@
 
 def confirm(
     message: str,
-    default: bool = True,
+    default: Optional[bool] = True,
     qmark: str = DEFAULT_QUESTION_PREFIX,
     style: Optional[Style] = None,
     auto_enter: bool = True,
     instruction: Optional[str] = None,
+    mandatory: bool = True,
```
spmvoss, line 26: I would strongly suggest defaulting `mandatory` to `False` to preserve the default behaviour and avoid breaking people's code. 

That will also resolve the issue with deletion of test cases. When mandatory defaults to false all existing test cases can remain the same and you only have to add new ones for mandatory=True case

PR #427, opened 2025-01-28
Fix dynamically-generated select choices

JamesGuthrie:

Dynamically-generated select choices were broken.

Example:

```python
import questionary
OPTIONS = {
    "key1": ["k1v1", "k1v2"], 
    "key2": ["k2v1", "k2v2"]
}

questions = [
    {
        'type': 'select',
        'name': 'key',
        'message': "Choose a key",
        'choices': OPTIONS.keys(),
    },
    {
        'type': 'select',
        'name': 'value',
        'message': 'Choose a value',
        'choices': lambda x: OPTIONS[x['key']],
    }
]
questionary.prompt(questions)
```

Results in the following error:

```
? Choose a key key1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../venv/lib/python3.10/site-packages/questionary/prompt.py", line 78, in prompt
    return unsafe_prompt(questions, answers, patch_stdout, true_color, **kwargs)
  File ".../venv/lib/python3.10/site-packages/questionary/prompt.py", line 219, in unsafe_prompt
    question = create_question_func(**_kwargs)
  File ".../venv/lib/python3.10/site-packages/questionary/prompts/select.py", line 149, in select
    if choices is None or len(choices) == 0:
TypeError: object of type 'function' has no len()
>>>
```

**How did you solve it?**

Modify the correct object.

**Checklist**

### DIFF
--- a/examples/dependent_selects.py
+++ b/examples/dependent_selects.py
@@ -1,0 +1,27 @@
+from pprint import pprint
+
+from questionary import prompt
+
+OPTIONS = {"key1": ["k1v1", "k1v2"], "key2": ["k2v1", "k2v2"]}
+
+
+def ask_dictstyle(**kwargs):
+    questions = [
+        {
+            "type": "select",
+            "name": "key",
+            "message": "Choose a key",
+            "choices": OPTIONS.keys(),
+        },
+        {
+            "type": "select",
+            "name": "value",
+            "message": "Choose a value",
+            "choices": lambda x: OPTIONS[x["key"]],
+        },
+    ]
+    return prompt(questions, **kwargs)
+
+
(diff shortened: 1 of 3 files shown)


kiancross:
Thanks for the PR! Do you think you'd be able to add a test to catch this bug? If not, no worries - I'll try to add one in the next two or three weeks before I merge this.

JamesGuthrie:
@kiancross I've added an example, and a test of the example.

JamesGuthrie:
@kiancross could you take another look at this PR?

tmbo (approved):
lgtm thanks a lot for the fix 🔥

PR #445, opened 2025-05-31
Fix default instruction

python-and-novella:
Fix default instruction,
The right is `(Use arrow keys to move, <space> to select, <ctrl-a> to toggle, <ctrl-i> to i
nvert, type to filter)`.


tmbo:
Thanks for the correction

PR #453, opened 2025-08-14
feat(): allow specific the initial_choice

JunchengXue:
Support to specific the initial choice 

...

**How did you solve it?**

...

**Checklist**


kiancross:
There are a few issues with this PR (unrelated changes due to f25a001de9ff1ee3034498b0941d6083c4b93272, the type hint, etc.), but before I go into depth, could you describe the motivation for it? `select` already starts the cursor on `default`, so I'm currently not seeing what `initial_choice` adds.

I'll close the PR in the meantime since I can't see the intent, but happy to reopen if you think there's a use case.

PR #498, opened 2026-05-13
feat(path): expose min_input_len for path autocompletion

### LINKED ISSUE #359: `path` parameter for `min_input_len`
### Describe the problem

I've found that I get 20-30x better performance (on Windows) if I I override the default completer for `path` prompts with the somewhat undocumented alternative of `completer=GreatUXPathCompleter(min_input_len=4)`. 

### Describe the solution

Any chance that `min_input_len` get added as a one of the standard inputs to `path` prompts? If not, I'd suggest adding the `GreatUXPathCompleter` solution to the docs.

### Alternatives considered

_No response_

SAY-5:

Closes #359. `GreatUXPathCompleter` already accepts `min_input_len`, but the public `path()` prompt didn't expose it, so users had to construct their own completer to delay autocompletion (which reportedly gives a 20-30x speed up on Windows for slow filesystems).

**How did you solve it?**

Added a `min_input_len: int = 0` keyword argument to `questionary.path()` that is forwarded to the default `GreatUXPathCompleter`. Default is `0` so existing behaviour is unchanged.

**Checklist**

### DIFF
--- a/questionary/prompts/path.py
+++ b/questionary/prompts/path.py
@@ -124,8 +124,8 @@ def path(
     only_directories: bool = False,
     get_paths: Optional[Callable[[], List[str]]] = None,
     file_filter: Optional[Callable[[str], bool]] = None,
-    min_input_len: int = 0,
     complete_style: CompleteStyle = CompleteStyle.MULTI_COLUMN,
+    min_input_len: int = 0,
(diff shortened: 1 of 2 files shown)


kiancross (approved):
Thanks for the PR!

PR #501, opened 2026-05-26
Add theming and styling reference.

Josverl:
**This PR adds information on theming**

I have found a need to improve on the default theming, and have not been able to find a clear source of information on which tokens affect what parts of which question type.
That has caused me some frustration due to the time needed to get a consistent look that works well across different terminals and dark/light preferences.

**How did you solve it?**

Used several AI passess to research the code base in this repo and prompt_toolkit in order to extract the most relevant tokens for each question type.
Then went through several manual passes to improve the structure of the presented information.

Note this is intentionally not cross-referenced with the other documentation - currently this is just a leaf-page connected from the index.

**Checklist**

- [x] I used AI tools to gather the information  presented here, but did review and hand edit it, and have done my best to verify its correctness.

### DIFF
--- a/docs/pages/theming_reference.rst
+++ b/docs/pages/theming_reference.rst
@@ -1,0 +1,1227 @@
+.. _theming-styling-reference:
+
+*******************************
+Theming & Styling Reference
+*******************************
+
+This  reference documents all style tokens used by questionary prompts, including which tokens each question type uses and how to override them.
+This is in addition to the overview provided in :ref:`advanced-themes-styling`.
+
+This page is up to date for questionary version : 2.1.1.
+Later / future versions may introduce new tokens or change defaults, so you may need check the source code for the most current information.
+
+Overview
+########
+
+What are styles?
+****************
+
+In questionary, styles are ``prompt_toolkit.styles.Style`` objects created from a list of 2-tuples like ``[("qmark", "fg:#673ab7 bold"), ("question", "bold")]``. Each tuple maps a **token name** (the style class) to a **prompt_toolkit style string** (colors, bold, italic, etc.).
+
+How styles work
+***************
+
+When rendering a prompt, questionary emits formatted text with class names like ``("class:qmark", "?")`` or ``("class:answer", "user input")``. The Style object maps these class names to visual styles. Questionary uses the prefix ``"class:"`` internally, but you define styles without it: just ``("qmark", "...")``.
+
+Style application
+*****************
+
+- Every prompt function accepts a ``style=`` parameter
+- Your custom style is merged with ``DEFAULT_STYLE`` using ``merge_styles_default()`` (source: ``questionary/styles.py``)
+- Custom styles **override** matching tokens in the default style
+- Merging order: ``DEFAULT_STYLE`` → user style → prompt_toolkit merge
+
+Recognized token names
+**********************
+
+**Questionary-specific tokens** recognized across questionary prompts:
+
+``qmark``, ``question``, ``answer``, ``instruction``, ``pointer``, ``highlighted``, ``selected``, ``separator``, ``text``, ``disabled``, ``search_success``, ``search_none``, ``question-mark``, ``validation-toolbar``, ``bottom-toolbar``
+
+**Prompt_toolkit tokens** inherited from the underlying UI framework (see the "Prompt Toolkit Style Tokens" section for complete list):
+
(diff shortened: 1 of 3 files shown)


docs/pages/theming_reference.rst
```diff
@@ -0,0 +1,1221 @@
...
+.. code-block:: python3
+
+   questionary.print("Hello!", style="fg:#00ff00 bold italic")
+
+form
+****
+
+**What renders:** Sequential execution of multiple question prompts. Not a unique prompt type - just a container.
+
+****Token usage:**** None directly. Each child question uses its own style.
```
kiancross:
```suggestion
**Token usage:** None directly. Each child question uses its own style.
```

docs/pages/theming_reference.rst
```diff
@@ -0,0 +1,1221 @@
...
+     - Search text when no matches
+
+rawselect
+*********
+
+**What renders:** Same as ``select`` but with mandatory keyboard shortcuts (``1)``, ``2)``, etc.) and no arrow key navigation.
+
+**Implementation:** Calls ``select(..., use_shortcuts=True, use_arrow_keys=False)`` (source: ``questionary/prompts/rawselect.py:69-79``)
+
+****Token usage:**** Identical to ``select`` above. All same tokens apply.
```
kiancross:
```suggestion
**Token usage:** Identical to ``select`` above. All same tokens apply.
```

docs/pages/theming_reference.rst
```diff
@@ -0,0 +1,1221 @@
...
+Toolbars: ``completion-toolbar``, ``search-toolbar``, ``system-toolbar``, ``arg-toolbar``
+
+UI elements: ``selected``, ``cursor-line``, ``cursor-column``, ``matching-bracket``, ``line-number``, ``auto-suggestion``, ``trailing-whitespace``, ``aborting``, ``exiting``
+
+You can style any of these tokens in your custom Style object.
+
+Default Style
+#############
+
+Defined in ``questionary/constants.py:37-57`` as ``DEFAULT_STYLE``:
```
kiancross: The line numbers throughout are at high-risk of drifting from the code.

docs/pages/theming_reference.rst
```diff
@@ -0,0 +1,1221 @@
+*******************************
```
kiancross, line 3: Should add a cross-reference to the styling section in `advanced.rst`.

kiancross (changes requested):
Thanks for this. I've left a few small inline suggestions.

My main hesitation is maintenance. This page restates a lot that lives in the source: the default style values, which tokens each prompt emits, and several hardcoded `file.py:line` citations, and nothing ties any of it back to the code. So as soon as a token is added, removed, renamed, or a default changes, this page becomes stale.

I don't think keeping 1200+ lines accurate by hand is a burden we can realistically sustain. It would be good to explore coupling it to the source instead. Some ideas:
 
- Generate the default-style table from constants.py:DEFAULT_STYLE at build time.
- Add a test that cross-checks the documented token names against the `class:` tokens actually emitted in `questionary/`, so add/remove/rename fails CI instead of rotting.

The optimal version would be a single token registry in code that both the runtime and these docs draw from, though that's a larger change than this PR.

What do you think?

docs/pages/theming_reference.rst
Josverl: Yes, fair point. I can probably just remove them all as they carry little value.

docs/pages/theming_reference.rst
Josverl: Makes sense, I'll add that.

Josverl:
I think maintenance is a fair point.

But honestly, if it was not for this document, I would not have been able to use your library. Some of the defaults are not legible enough. And I know some of the users of my tools appreciate that things look nice, and be clearly readable with sufficient contrast.
I just could not find that information in any documentation, I might have missed it.

A possible approach to keep the doc updated is to use the same tech that created it.
With GitHub agent actions it should be simple to have an AI agent update this doc if either the source, or the pinned dependency for prompt library changes.

Then with a human review, I think the results will be better than any script output.

I do not intend to spend time on creating custom scripts or learning the rather extended codebase well enough to pull that off.  I have 40+ weekend project on my post it scrum board as it is.

I understand if you don't merge this to avoid the maintenance burden.

Josverl:
updated as requested,

Josverl:
ill close this in about a week from now if there is no interest.

PR #503, opened 2026-05-29
fix: don't render a select's default as a selected item

koriyoshi2041:
Fixes #473.

In a single `select`, the `default` choice is rendered permanently with the inverted "selected" style, even when the cursor is on a different choice.

Root cause: `select()` passes `default` to `InquirerControl` twice — positionally **and** as `initial_choice`:

```python
ic = InquirerControl(
    choices,
    default,            # -> self.default -> _is_selected() -> selected_options
    ...
    initial_choice=default,   # -> pointer position
)
```

The positional `default` ends up in `selected_options`, which the renderer styles with `class:selected` regardless of the pointer position. But a single select has no persistent selection — only a pointer. `initial_choice=default` already places (and validates) the pointer, so the positional `default` only causes the buggy styling. The docstring also describes `default` solely as setting the pointer position.

Fix: drop the positional `default` from the `InquirerControl(...)` call, keeping `initial_choice=default`. (`checkbox` legitimately uses `selected_options`, so `_is_selected`/`common.py` are untouched.)

Added `test_select_default_not_rendered_as_selected`, which builds a `select` with a default and asserts the control's `selected_options` is empty while the pointer is on the default. It fails on `master` (selected_options contains the default) and passes with the fix; all existing `test_select.py` tests still pass.

### DIFF
--- a/questionary/prompts/select.py
+++ b/questionary/prompts/select.py
@@ -164,7 +164,6 @@ def select(
 
     ic = InquirerControl(
         choices,
-        default,
         pointer=pointer,
         use_indicator=use_indicator,
         use_shortcuts=use_shortcuts,
--- a/tests/prompts/test_select.py
+++ b/tests/prompts/test_select.py
@@ -4,7 +4,9 @@ from copy import copy
 import pytest
 
 from questionary import Choice
+from questionary import select
 from questionary import Separator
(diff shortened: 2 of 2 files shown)


kiancross (approved):
Thanks for the PR!

